PR #227, opened 2015-04-04
Fix output of python example in Readme.

debugger22:


msoos:
Thanks, nice catch!

msoos:
Also, there was another bug, but clause to be added was not [-3] but [3] or we would never get a SAT solution at the first place. Fixed that too :)

debugger22:
Oh yes!. I got `sat == False ` but then I thought I must have missed something. :)

PR #229, opened 2015-04-13
Fix CMake for local install m4ri

### LINKED ISSUE #228: Compile error: Mismatch between FindM4RI.cmake and xorfind.cpp
I try to build [cryptominisat 4.2.0](http://msoos.org/largefiles/cryptominisat-4.2.0.tar.bz2) with [m4ri-20140914](http://m4ri.sagemath.org/downloads/m4ri/m4ri-20140914.tar.gz) on Linux (Fedora 20, CMake 2.8.12.2). I manually installed m4ri to a local directory (`/tmp/m4ri-20140914`). m4ri is otherwise not installed on the system.
 
I build cryptominisat and get a compile error:
```
$ mkdir build && cd build
$ export CMAKE_PREFIX_PATH="/tmp/m4ri-20140914"
$ cmake ../cryptominisat4-4.2.0
$ make
...
.../cryptominisat4/xorfinder.cpp:28:23: fatal error: m4ri/m4ri.h: No such file or directory
 #include <m4ri/m4ri.h>
```

I could Identify the reason to a mismatch of FindM4RI.cmake, include_directories and xorfinder.cpp.

In [FINDM4RI.cmake](https://github.com/msoos/cryptominisat/blob/master/cmake/FindM4RI.cmake#L36) `M4RI_INCLUDE_DIRS` is set to `/tmp/m4ri-20140914/include/m4ri` but  [xorfinder.cpp](https://github.com/msoos/cryptominisat/blob/master/cryptominisat4/xorfinder.cpp#L28) includes `m4ri/m4ri.h` which can not be found in any include directory. I assume this problem is not visible on systems that have m4ri installed system-wide as `/usr/include` contains `m4ri/m4ri.h`.

I think the solution is to change `xorfinder.cpp` to include `m4ri.h` as the pkg-config file for m4ri also uses the include/m4ri directory:
```
# from lib/pkgconfig/m4ri.pc
includedir=${prefix}/include
Cflags: -I${includedir}/m4ri -g -O2  -mmmx -msse -msse2 -msse3 
```

My workaround is to overwrite the cache variable  `M4RI_INCLUDE_DIRS`:
```
  $ cmake -DM4RI_INCLUDE_DIRS="/tmp/m4ri-20140914/include" ../cryptominisat4-4.2.0
```

finnhaedicke:
Revert the previous fix, change FindM4RI to use `.../include` instead of `.../include/m4ri`.

fixes #228


msoos:
Thanks! :)

PR #231, opened 2015-04-14
Fix compilation on OS X

rion:
Removes the following errors when compiling on OS X:

/Users/rion/cryptominisat/cryptominisat4/bva.cpp:182:66: error: no member named
      'sqrt' in namespace 'std'; did you mean simply 'sqrt'?
    *simplifier->limit_to_decrease -= 2*(long)m_cls.size()*(long)std::sqrt(m...
                                                                 ^~~~~~~~~
                                                                 sqrt

/Users/rion/cryptominisat/cryptominisat4/bva.cpp:670:76: error: no member named
      'log' in namespace 'std'; did you mean simply 'log'?
  ...-= (double)potential.size()*(double)std::log(potential.size())*0.2;
                                         ^~~~~~~~
                                         log

### DIFF
--- a/cryptominisat4/bva.cpp
+++ b/cryptominisat4/bva.cpp
@@ -25,6 +25,7 @@
 #include "clausecleaner.h"
 #include "subsumeimplicit.h"
 #include "sqlstats.h"
+#include <cmath>
 #include <functional>
 
 using namespace CMSat;


msoos:
Good stuff, thanks! So it compiles other than this? I'm surprised! TravisCI has an experimental feature for MacOS X support, I'll look into it so you'll have a more smooth experience next time!

Thanks again,

Mate

PR #232, opened 2015-04-15
link public against M4RI_LIBRARIES

finnhaedicke:
This way other CMake projects can pick up M4RI, even when it is installed locally.

Currently the imported target `m4ri` is exported, which will resolve to a locally installed m4ri library in `/usr/lib`, but will not work when m4ri is installed locally.
With this patch the full path to the library is exported.

This works, when m4ri, cryptominisat and the client using it (e.g. STP) are build on the same system but will fail when cryptominisat is packaged and used on a system with a different location. However currently this also only works when m4ri is installed system-wide on the new system.

### DIFF
--- a/cryptominisat4/CMakeLists.txt
+++ b/cryptominisat4/CMakeLists.txt
@@ -52,16 +52,8 @@ set(cryptoms_lib_link_libs "")
 if (M4RI_FOUND)
     include_directories(${M4RI_INCLUDE_DIRS})
 
-    if (STATICCOMPILE)
-        add_library(m4ri STATIC IMPORTED)
-    else (STATICCOMPILE)
-        add_library(m4ri SHARED IMPORTED)
-    endif (STATICCOMPILE)
-
-    set_property(TARGET m4ri PROPERTY IMPORTED_LOCATION ${M4RI_LIBRARIES})
-
     SET(cryptoms_lib_files ${cryptoms_lib_files} xorfinder.cpp)
-    SET(cryptoms_lib_link_libs ${cryptoms_lib_link_libs} m4ri)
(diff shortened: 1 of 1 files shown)


finnhaedicke:
Two more comments:
1) The user of cryptominisat still has to set LD_LIBRARY_PATH for m4ri (except in the CMake build dir, where cmake sets the RPATH) 

2) You can see how we integrated cryptominsat in our dependencies repository here: https://github.com/finnhaedicke/dependencies/tree/preview-stp-git/cryptominisat-git

msoos:
Thanks for the above! Nice one!

As for RPATH... hmmm. Should be updated after install to the correct path, I thought. There are ways of doing it I think. It's a common issue: http://www.cmake.org/Wiki/CMake_RPATH_handling I'll take a look at it, maybe I can figure something out.

finnhaedicke:
I already implemented something like this for metaSMT: https://github.com/finnhaedicke/metaSMT/blob/development/cmake/library_rpath.cmake

It generates some warnings with newer versions of CMake, but maybe I can have another look at it.

finnhaedicke:
I pushed a fixed version of library_rpath.cmake in https://github.com/agra-uni-bremen/metaSMT/commit/be176cc632777a4a353b6cef2ec30709978d5194. Maybe you want to have a look at it.

msoos:
OK, thanks, I'll see if this could be used by my setup. In the weekend hopefully I'll have time :)

msoos:
Oh, and I just looked at your work, http://www.informatik.uni-bremen.de/agra/doc/konf/2014_dvcon_europe_crave2.pdf looks really great. I'll take a look a deeper look at CRAVE (amazing template trickery!) and see how you use SMT. If you have any questions regarding SMT/STP/SAT/MiniSat/etc don't hesitate to contact me!

PR #253, opened 2015-09-12
build: stricter GCC flags.

aodinaev:
- treat some GCC warnings as errors
 - fixed some warnings


msoos:
Hmm... I like this stuff! But I need you to split this up to a number of different patches :) I commented on the patch itself, please take a look!

aodinaev:
I suggest to create separate patch with ```CMakeLists.txt``` changes and whatever it implies since ```Werror=``` will prevent project to compile. 
Comment all ```watch_idx_t``` fixes as upcoming feature, not sure about ```xor_t``` though.

msoos:
Oh, on that note... Werror is very dangerous on release. It means that if we release with Werror, and gcc improves and finds another thing that is unused, future compiles will break for the users of old cryptominisat versions. I have hit myself with this before and so I will never do Werror on anything that's released. We can have an "INTERNAL_BUILD" cmake flag that we always set that sets Werror, though, and I'm happy with that.

msoos:
In the end, a lot of this has been merged I believe in another pull request. So, I'm closing this :)

PR #254, opened 2015-09-12
watched: streamlined getType() method.

aodinaev:
- used static_cast<> instead of chain of if/else's.
 - initialized type member explicitly in default constructor.


msoos:
Thanks!

PR #255, opened 2015-09-13
warnings: suppressed watch_idx_t warning.

aodinaev:
Extracted ```watch_idx_t``` related changes. I skipped ```Werror``` CMakeLists.txt changes for now.

### DIFF
--- a/src/bva.cpp
+++ b/src/bva.cpp
@@ -174,6 +174,10 @@ void BVA::remove_duplicates_from_m_cls()
                     }
                     return false;
                 }
+                case CMSat::watch_idx_t: {
+                    // entry for upcoming feature to suppress compiler warning.
+                    break;
+                }
             }
 
             assert(false);
@@ -223,6 +227,11 @@ void BVA::remove_duplicates_from_m_cls()
                         break;
                     }
                 }
+                break;
+            }
+
+            case CMSat::watch_idx_t: {
+                // entry for upcoming feature to suppress compiler warning.
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


msoos:
Ah, actually, idx_t will never be in bva.cpp. So I fixed that. Also, I _never_ set default values for something that will later be set. This way, valgrind will warn me if something is off. I don't mind the gcc warnings, they are stupid. I want the real error, from valgrind to be displayed in all red :)

aodinaev:
Not trying to teach you how to run your project, but imho, these are weak arguments vs strong type checking which C++ compilers provide.
Valgrind, in opposite of C++ compiler, is path dependent tool and may or may not cover all possible scenarios. Quality of valgrind checks depends on test coverage, which is by nature of computation can never be 100%.
Also, you can't require all developers to run valgrind to check errors. In long run, it will become slow and annoying process.
Imagine situation where it takes couple hours to detect trivial uninitialized variable.

Usually developers prefer earliest possible notifications when something is off. 
First, compiler errors. Obviously you don't want to deal with code which can't be compiled
Then warnings. Compiler output should be warning free. Reason behind it, one doesn't need to maintain mental map which warning is ok and which is not. Also, in long run, you run a risk when useful warning can be lost in other warnings noise. Paranoid warning are dealt explicitly either by compiler settings or locally with pragmas if there is no way around it.
And only when compiler yields clean output, Unit tests, Regression tests, Valgrind and other sanity check tools.

As I said, I'm not trying to teach you how to run the project, but please imagine that you are the developer who wants to contribute to cryptominisat4. Currently you would need to go through each warning message and make sure nothing important hiding behind each of them.

msoos:
First of all, thanks for the feedback!

The problem with setting x to any value is that it's inherently wrong, whatever value it has. It should never execute with that default value. So, setting it to anything doesn't really help -- it's just as wrong as if it wasn't set. The advantage, is, though, is that valgrind will catch the version where it's not set :) But I agree that it's annoying to have that compiler warning. Let's try to keep that one in. The problem is really the others, that shouldn't be there...

Actually, valgrind is pretty fast if you run it with the fuzz setup I have and that you can have too :) Please check out the .travis.yml and the travis-cmake.sh files. They detail how to set it up (git submodules, build, then execute scripts/fuzz_test.py). These fuzzers are running at every push to the git. Note that I couldn't make valgrind run on TravisCI :( But it runs fine on my system. Also, the coverage is remarkably high. Really, check it out. 

I agree that compiler output should be warning free. Lately I haven't been doing that too much :S My fault, sorry. Thanks for cleaning so much of them up!

Yep, I agree with what you are saying in 99% of the cases, except for that x=false :) Most other places I tried to set default values if they make sense (they almost always do).

Again, thanks for the feedback! If you have any other diffs, I'd be happy to merge them. Also, I'll go through the code in the coming days and try to fix the warnings. Note that different compiler versions give different warnings(!) so if I don't fix all your warnings that might mean that you simply have a different compiler version :( Anyway, I'll try :) Thanks again!

Mate

PR #258, opened 2015-09-14
Fix/prop engine

aodinaev:
very experimental. Goal of this PR is to start discussion how branch mispredictions can be avoided at large.

so far, I wasn't able to capture significant performance improvement related to this PR. Though, I have got impression that ```propagate_any_order``` function is being used intensively during late stages.

Please let me know if there is better way to contrast new and old solutions.


msoos:
First of all, thanks!

Just a quick note about the "merge branch" commit. If you do "git fetch" then "git rebase" (instead of "git pull") then your  commits will be rebased on top of the changed system and you will not have to do the merge. I try to avoid these merge commits like the plague :) Also, you can simply not merge and request a pull request for only the change you did. Anyway, this is kind of minor. I would normally ask you to give a pull request with only the 1st commit in it (should be easy, I think!)

Okay, so on to the more interesting part :) Sounds cool, but I wonder if such a function pointer structure is faster. It certainly looks cleaner, but this is the part of the system that is _highly_ critical in terms of speed, and any indirection can cause a 10+% slowdown, which is quite a lot for purely cosmetic reasons. Have you checked that applying commit 1 indeed improves speed? Something long-running like 600s or longer would be good to see. Using a longer problem like that should make it obvious if the change speeds things up. Do you have any such benchmarks? Would be good to see some numbers!

aodinaev:
merging branch into original branch is being me lazy. I didn't want this branch to be published but decided to let you evaluate it.

Speaking of original change. Goal was to accelerate ```propagate_any_order()```. it seems it didn't improve. though I gained from eliminating branch mispredictions I lost on inlining (before methods were inlined, with indirection compiler couldn't optimize this way).
I will experiment more on weekend but don't expect too much.

msoos:
:) I suggest you take a look at other issues... Martin Maurer just found out that when finding SAT solutions, the threads that didn't find the solution didn't exit immediately :S So were wasting time for them to find the solution too. Pretty crazy bug :( There must be others too. Anyway, I am writing a test for this one during the weekend so we don't have one like this again :) Thanks in advance for your work on CryptoMiniSat!

PR #263, opened 2015-09-16
Add option to disable Python in CMake build

bmastenbrook:
I had trouble compiling the Python interface in my setup (it seemed to pick up a compile flag from the system Python that the clang I was using did not support). Since I'm not using Python, I ended up adding an option to the CMake build to disable it, which might be useful to other people as well.


msoos:
Strange. It should have found that python is not there and simply not compiled it. I'm puzzled. Do you know what could be done to make sure that the tests are better?

msoos:
(note that I merged anyway, as I see no harm in merging, thanks!  :) )

PR #275, opened 2015-10-07
Fix interruption condition in handle_sub_str_with

4tXJ7f:
Fix the issue reported by Clang:

```
warning: & has lower precedence than ==; == will be evaluated first
[-Wparentheses]
```


msoos:
Thanks! Good catch :)

PR #276, opened 2015-10-07
Fix unreachable code warnings

4tXJ7f:
When not using flags like DEBUG_RESOLV, some functions do nothing. This
was achieved by having an unconditional return in an `#ifndef`
statement. The return statement, however, makes the rest of the function
(intentionally) unreachable and triggers a warning when compiling with
clang. The solution is to restructure the affected functions slightly to
avoid these warnings.

### DIFF
--- a/src/cnf.cpp
+++ b/src/cnf.cpp
@@ -449,10 +449,7 @@ void CNF::load_state(SimpleInFile& f)
 
 void CNF::test_all_clause_attached() const
 {
-#ifndef DEBUG_ATTACH_MORE
-    return;
-#endif
-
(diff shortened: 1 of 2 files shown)


msoos:
Thanks! Though the avoid-unused-param-warning is such a massive hack. I don't like that kind of thing. Compiler warnings are meant to make the code better, not worse. I might undo them or just change it. It's utterly ridiculous, that it is just code mess, which is hard to maintain.

4tXJ7f:
Yeah, agreed :) The warning should somehow take into account that the parameter is used when compiled with certain options.

PR #277, opened 2015-10-07
Add missing include

4tXJ7f:
Looks like commit 5a40d1b9c3053c14e1f4bfd6b92326b415ebdfec broke the
build on clang by indirectly removing math.h from the included header files.


msoos:
Thanks, great work!

PR #278, opened 2015-10-08
Fix python build on OSX

4tXJ7f:
Use python-config to get the correct parameters for building/linking.
Limitation: flags returned by python-config cannot contain spaces
because spaces are used to split the flags.


msoos:
Thanks!

msoos:
Note 6d02f06f9b44a3462a18f83e531ff47a5948d27b It's weird as hell, but c++ was called with the parameters passed using argc+argv and so an empty '' element in the array meant a file called '' (i.e. empty filename) was to be compiled. c++ didn't find the file and died, telling me an error message that didn't have a filename inside. Anyway, fixed! :)

4tXJ7f:
Hm interesting, good to know :)

PR #279, opened 2015-10-08
Only attempt coveralls upload if $COVERTOKEN set

4tXJ7f:
Travis fails for pull requests because CMS_CONFIG=COVERAGE uses
$COVERTOKEN when uploading results to coveralls and $COVERTOKEN is not set.
Check if $COVERTOKEN is set and skip the result upload if it is not.


msoos:
Thanks. Makes a lot of sense :)

PR #281, opened 2015-10-09
Fix while loop condition in skipWhitespace()

4tXJ7f:
Commit f10db78e23dd180abcab74f4e7d1beef99035e89 changed the while loop
condition slightly (accepting '\n' instead of '\t'), leading to failing
tests.


msoos:
Sorry, was working on fixing it in the meanwhile :)
Thanks!

PR #286, opened 2015-10-15
fixed bool_switch and short_options

a1880:


a1880:
Rewrote argc/argv scanning to avoid intermediate data structures and issues with Berkeley comments. It would be fair to call it cheating, but some parts are clearer and easier to read now.

Under Windows 7, departure from Boost reduced the size of the cms4.5.3 executable roughly from 1.5 to 1.3mb. The result looks rather fast.

Still, some additional testing would make sense before merging back into the master branch.

msoos:
Thanks! I am currently in the middle of another thing, I'll look at this tomorrow! Sorry again for the late replies. I am trying my best have have been working CMS this the past hours. Just not on this part. The builds and checks are failing everywhere so I need to be a bit more on top of things. Sorry about that!

PR #291, opened 2015-10-25
removed old entry in CMakeLists.txt

a1880:


msoos:
Thanks! This might fix all the issues with Travis CI :) I don't know if you have been getting the mails. It's pretty useful actually, tells you when things are not 100%.

PR #319, opened 2016-02-24
Interface extension - no_bva

SmallLars:
I would like to add this function to the interface, because its not possible to use the open_file_and_dump functions without disable bva.


msoos:
Thanks, looking great!

PR #329, opened 2016-04-08
Kuldeep new

kuldeepmeel:
I have added ScalMC Algorithm code and option to specify which algorithm should CUSP choose.


msoos:
Hey Kuldeep. As discussed before, I have no time to do this, there is the SAT competition deadline. However, may I ask you to put ScalMC into a different file entirely, with its own executable? Essentially, copy cusp.cpp into scalmc.cpp and put all your changes there. We need to completely separate these algorithms, at least at the beginning. I suspect there will be a lot to clean up in the ScalMC code as there was with CUSP. I don't want to loose all the work I have already done on CUSP. So, can you please put ScalMC into a completely different file? Thanks a lot!

Issue #320, opened 2016-02-24
Adding interface functions do help CMS gaining knowledge while running?


SmallLars:
Hello,

I am working on my master thesis and using CMS (4.5.3) to analyse SHA256. Therefore I do different runs and try to evaluate new clauses I get from different dumps. After that I try to multiply possibly valuable clauses in 2 directions: rounds (64) and bit width (32). With this strategy I was able to find a lot of more possibly valuable clauses. At this time I do this steps by hand and thinking about a better solution:

I think its not a good way to extend CMS in a direct way to do this work, because the problem is very 
specific. But how about an extension of the interface in a general way, to make it possible to use external knowledge? I think about an interface function, where I can add a function pointer to CMS.
Every time CMS finds a new clause (redundant, irredundant or assumption based) CMS call this function to ask what to do. It should be possible to throw the clause away, confirm usage or add more clauses to CMS. With this method it would also be possible, to try new evaluation strategies for clauses, without touching the internal code of CMS or adding a lot of new options.

I would be happy when you can tell me what you think about this, before I start a lot of useless work.
Is it possible to implement this in a reasonable easy way? Would you be interested to add this on CMS or merge a possible pull request?

Thank you very much
   Lars

msoos:
Sounds possible and not horrible. However, note that your function will have to run _very_ fast. It will be called >20'000 times a second. Throwing away the clause is not really possible, but we can set its "glue" very high and so it will not be retained for long. Adding extra clauses through that callback function will be _very_ tricky. Making the clause permanent is easy, though, you can again manipulate its glue, setting it low, and thereby not removing it.

Here is where the callback should be and you should set the glue here:


```
    minimize_learnt_clause();
    glue = calc_glue_using_seen2(learnt_clause);
    mimimize_learnt_clause_more_maybe(glue);
```

AFTER the `glue = ...` you should put your function call, and basically do:

```
glue = my_function(learnt_clause)
```

Return a value of 2 in case you want to keep it, and return a value of 50 in case you'd rather not.

rw:
In a similar way, is it possible to customize the order in which variables (CNF inputs) are selected for evaluation?

msoos:
Sure. Look at the `pickBranchLit` function :) Note that variables are renumbered internally (kinda randomly), so you probably want to pass `--renumber 0`. Good luck!

rw:
@msoos Great, thanks!

msoos:
[...]

PR #336, opened 2016-05-11
Interface extension for glue manipulation

SmallLars:
As in #320 suggested I added code to add a function for glue manipulation without changing CMS code. There is an example for usage below. Maybe you like to merge.
```
uint32_t manipulate(vector<Lit>& clause, uint32_t glue) {
    // Accept and pass glue or change it
    return glue;
}
solver.append_glue_calc(manipulate);
```

### DIFF
--- a/src/searcher.cpp
+++ b/src/searcher.cpp
@@ -763,6 +763,9 @@ Clause* Searcher::analyze_conflict(
         }
     }
     glue = calc_glue(learnt_clause);
+    if (conf.calc_glue_user != NULL) {
+      glue = conf.calc_glue_user(learnt_clause, glue);
+    }
     print_fully_minimized_learnt_clause();
 
     if (learnt_clause.size() > conf.max_size_more_minim
--- a/src/cryptominisat.cpp
+++ b/src/cryptominisat.cpp
@@ -418,6 +418,16 @@ DLL_PUBLIC void SATSolver::set_no_bva()
     }
 }
 
+DLL_PUBLIC void SATSolver::append_glue_calc(uint32_t (*function)(vector<Lit>&, uint32_t))
+{
+    for (size_t i = 0; i < data->solvers.size(); ++i) {
+        Solver& s = *data->solvers[i];
+        s.conf.do_bva = false;
+        s.conf.doRenumberVars = false;
+        s.conf.calc_glue_user = function;
+    }
+}
+
 DLL_PUBLIC void SATSolver::set_verbosity(unsigned verbosity)
 {
   for (size_t i = 0; i < data->solvers.size(); ++i) {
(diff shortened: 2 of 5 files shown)


src/searcher.cpp
```diff
@@ -756,13 +756,12 @@ Clause* Searcher::analyze_conflict(
     stats.litsRedNonMin += learnt_clause.size();
     minimize_learnt_clause<update_bogoprops>();
     stats.litsRedFinal += learnt_clause.size();
+    glue = calc_glue(learnt_clause);
     if (learnt_clause.size() <= conf.max_size_more_minim) {
-        glue = calc_glue(learnt_clause);
         if (glue <= conf.max_glue_more_minim) {
             minimize_using_permdiff();
```
msoos: After this line, you need to re-calculate the glue. It might have changed, since minimization can reduce the size of the clause, thereby changing the glue in it.

msoos:
What I'm a bit surprised about is that the function will have access to only the literals inside the clause. I'm not entirely sure that will help as you probably want to have access to much more that than that. At the very least, the renaming scheme, since variables are renamed internally -- or you need to make sure that variables are never renamed when this is in use. Most people would not realize that, though and have trouble using this API. So renaming datastruct is essential. So I'm afraid this will not really bring too much benefits. Can you please tell me what you want to use this for?

SmallLars:
Thank you for your comments. I only need access to the literals inside the clause. For every literal I am able to find the position where it was used. Based on this information I calculate a "distance" in an undirected graph for every clause. I sorted many clauses by hand and was able to show, that clauses with a distance >= 0 speed up solving process. Most clauses have a distance < 0 and slow down solving process. So what I want to do is to keep clause amount low marking clauses with distance < 0 with a glue of 50. In addition I want to check if clauses with distance >= 0 have a low glue and correct it if its not. The focus of this is to optimize/extend the CNF for future solving process without glue correction and also collect additional information for the problem.

I think with disabled BVA there is no variable renaming so I disabled BVA when a user specific glue function is used. For me its not important that you merge this pull request. I only want to show you how I use CMS so you are able to pick up that idea. Maybe there is a better way to provide external knowledge to CMS while running?

capiman:
Perhaps 

  --renumber arg (=1)      Renumber variables to increase CPU cache efficiency

used as

--renumber=0

is also needed?

msoos:
@capiman is exactly right. So I suggest you add that too, and then re-do the whole experiment and check that your calculations still hold. Have you tried it yet by the way? How much more instances does it solve, with e.g. the SAT Competition 2015 instances, over a 3000s of timeout? Or how much faster is it on average on the instances it does solve? It's not trivial to conduct do this kind of experiment and it's needed for proper understanding of the benefits. I'd be really-really curious!

PR #355, opened 2016-07-17
Fix building the static version of cryptominisat

MartinNowack:
Hi @msoos I tried out the new cryptominisat, I had some issues linking against it.
Following is cryptominisat part:

* Do not link statically against pthread to avoid issues
  if (g)libc is upgraded
* Use CMAKE_THREAD_LIBS_INIT support
* simplify linking
* correctly export path of libcryptominisat5.a: allows to find it
  if it is installed in a different place
* Use LINK_SEARCH_START_STATIC to support static linking of executables
* Typo: Changed "No pthread" to "Pthread"


msoos:
I think there are some issues here, almost all of TravisCI failed. I'll look into it, I think there are some good ideas in there, but we need to be a bit more careful :)

By the way, you can always build with:

```
#!/bin/bash

set -e

cd build
rm -rf cm* CM* lib* cryptomini* Testing* tests* pycryptosat include tests
cmake -DENABLE_TESTING=ON -DREQUIRE_M4RI=ON -DSTATICCOMPILE=ON -DENABLE_PYTHON_INTERFACE=OFF -DNOVALGRIND=ON -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTING=OFF ..
make -j6 VERBOSE=1
```

Have you tried doing that?

MartinNowack:
@msoos I rolled back major changes as well, only fixing the build location and statically building some of the test applications.

Beside that, the submodule: utils/sha1-sat has to be fixed as well:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 18f6d1b..92f8af4 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,10 @@ find_package( Boost 1.46 REQUIRED COMPONENTS program_options)
 add_executable(sha1-gen
     main.cc
 )
+    if (STATICCOMPILE)
+        set_target_properties(sha1-gen PROPERTIES LINK_SEARCH_END_STATIC 1)
+        set_target_properties(sha1-gen PROPERTIES LINK_SEARCH_START_STATIC 0)
+    endif (STATICCOMPILE)

 target_link_libraries(sha1-gen
     ${Boost_LIBRARIES}
```

And what I just saw is that the in
https://github.com/msoos/cryptominisat_build/blob/master/build_static.sh
the `-DENABLE_TESTING=OFF` at the end should be removed.

Thanks for your time!

msoos:
Wow, great work, thanks!

Issue #359, opened 2016-07-21
scalmc doesn't generate new XORs after timeout


dfremont:
The code which is supposed to force the generation of a new set of XOR clauses after a timeout doesn't work (I confirmed with Kuldeep that, like in the old ApproxMC implementations, the desired behavior is to pick new hash functions after a timeout). It's a one-line fix: the `hashVars` map needs to be cleared as well as `assumps`. I don't know how to make an issue and a pull request simultaneously so I will make a pull request right after submitting this. I've tested the fix on a bunch of files and I think it is correct.

(FYI, the code that is supposed to increase the hash count after three timeouts also doesn't work, but it's not clear if that's what you want to do now that we're doing binary search, so I'll leave it to Kuldeep to think about.)

Let me just note that the motivation for doing this is that in our old experiments, you occasionally got a difficult set of XORs that took forever to solve; so timing out and starting over with a new set of XORs actually saved time, even though you lost some incrementality. I've done a few experiments using different random seeds for scalmc and it looks like this can still happen, although I haven't done anything rigorous. At any rate, if you switch from a fixed user-specified timeout to one that grows automatically, it might no longer make sense to always throw out the XORs after a timeout. But if the timeout stays fixed, it's important to pick new XORs.

msoos:
We still need to fix the random seed not being manipulable from the command line.

msoos:
OK, seed issue is fixed too.

PR #360, opened 2016-07-21
hashes actually regenerated on timeout

dfremont:
Fixing #359.


kuldeepmeel:
Thanks Daniel for the fix.

msoos:
Thanks! To be honest, we need proper testing for all of this. This fix fixes it once, but it might be broken again. I really want to add tests so things don't keep breaking :)

PR #364, opened 2016-08-16
Support for model counting under projection

wadoon:
This patch extends cryptominsat to support for model counting under projection.

This patch contains: 
* Modification of dimacsparser to support `cr [...]  0` lines
* Extension of `multi_solutions` for projection scope  

After the patch CM successfully counts 32 of following input: 

```
p cnf 10 0 
cr 1 2 3 4 5 0 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ scripts/fuzz/tmp_for_xor_to_cnf_*
 scripts/reconf/outfile*
 
 web/jquery-1.11.3.min.js
+build
--- a/src/cryptominisat.h.in
+++ b/src/cryptominisat.h.in
@@ -32,6 +32,7 @@ THE SOFTWARE.
 #include <utility>
 #include <atomic>
 #include "cryptominisat5/solvertypesmini.h"
+#include <set>
 
 namespace CMSat {
     struct CMSatPrivateData;
@@ -46,6 +47,7 @@ namespace CMSat {
         , std::atomic<bool>* interrupt_asap = NULL
         );
         ~SATSolver();
+        void set_num_threads(unsigned n);
         unsigned nVars() const;
         bool add_clause(const std::vector<Lit>& lits);
         bool add_xor_clause(const std::vector<unsigned>& vars, bool rhs);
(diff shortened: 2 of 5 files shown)


kuldeepmeel:
This patch is not required. CMS Is already doing counting under projection. Just use "c ind " instead of "cr ". This is something that has been known since 2014 (See: http://www.cs.rice.edu/~kgm2/Papers/DAC2014.pdf)

msoos:
First of all, thanks for contributing code!

This is an interesting one. The "c ind" is actually not used by "main.cpp". So @kuldeepmeel is right that "c ind" is there, but unfortunately, it's not used by main.cpp. So, let's do the following :) I manually merge the parts that deal with using the "c ind" inside main.cpp. That will achieve what this patch adds. I hope that works? I'll get back here with the exact diffs that achieved the intended results and close the issue. You can either then comment here or open an issue in the issue tracker! I'm actually **really** interested in your opinion. I'm just closing this so I know what needs to be done, i.e. purely for house-keeping.

Thanks again, it's well-spotted that it wasn't used in main.cpp. Fixing now :)

msoos:
Fixed with 804363da839979cf4636ad8ec9fdb5bfdb226e63 I also added tests, so it should be OK.

Please let me know if it works as intended. Here is a snippet:

```
c ind 1 0
1 2 3 0
-1 3 0
-1 2 0
2 3 0
-2 -3 0
```

Gives you exactly 1 solution in case you pass `--maxsol 20`. In case you remove the

```
c ind 1 0
```

It will give you 2 solutions. The independent variable above is indicated to be 1. To indicate more:

```
c ind 1 2 3 4 [...] 0
c ind 201 202 203 [...] 0
[...]
```

I hope it makes sense. Let me know if there are any issues. Please open a new issue here on GitHub and I'll fix it. Thanks again for the bug report!

wadoon:
Thank you! 

If you want to compatible to `dsharp`, `sharpSAT` or `clasp`, you can easily add support for `cr` with following patch:
```diff
@@ -338,7 +338,7 @@ bool DimacsParser<C>::parseComments(C& in, const std::string& str)
         if (verbosity >= 6) {
             cout << "c Parsed Solver::new_vars( " << n << " )" << endl;
         }
-    } else if (str == "ind") {
+    } else if (str == "ind" || str == "r") {
         if (!parseIndependentSet(in)) {
             return false;
         }
```

msoos:
Thanks, but currently, nobody I know uses the dsharp and "c r" syntax.

PR #368, opened 2016-09-12
Put -I flag for generated headers earlier in the command line.

arrowd:
Otherwise, compiler may pick headers from installed cmsat.


msoos:
Thanks, good stuff!

PR #370, opened 2016-09-20
fixed xor generation bug

kuldeepmeel:
Random XORs were not being correctly generated.


msoos:
Thank you Kuldeep for finding these

PR #377, opened 2016-12-21
Use GNUInstallDirs on Linux, Fixes #376

### LINKED ISSUE #376: CMake doesn't respect arch specific install directories
Hi,

For things like libraries, cryptominisat should install to lib/lib64 depending on architecture. An easier workaround is to use GNUInstallDirectories and fallback to bin/lib/include on other platforms.

ambasta:
Updated CMakeLists to use GNU standard installation directories on non-windows platform!


msoos:
OK, hand-merged. There were some issues, you can check them out. The https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html gives some ideas, basically, LIBDIR is not defined, only CMAKE_INSTALL_LIBDIR, etc.

ambasta:
Thanks!

Issue #369, opened 2016-09-15
Python 3 binding


tobiasBora:
Hello,

First thank you for your program. I would like to know, when I install cryptominisat with cmake it installs pycryptosat for Python2, but not for Python 3. Is there any reason not doing that ?

Thank you,

TobiasBora.

msoos:
Not really. I'd be really happy if you could fix that! Just let me know how you did it, e.g. by cloning the repository, pushing your changes, then asking for a pull request:

https://help.github.com/articles/about-pull-requests/

Looking forward to the pull request!

Mate

msoos:
So I guess this is fixed, right? I'm closing, but please re-open in case there are issues :) Thanks a LOT for the pull request, really great work!

PR #382, opened 2017-03-08
Python3

sliedes:
Python 3 bindings (Issue #369). Also in preparation for that silence a couple of warnings and fix a couple of typos in python/README.rst.

### DIFF
--- a/python/README.rst
+++ b/python/README.rst
@@ -108,8 +108,8 @@ absolute value corresponds to i\ :sup:`th` variable::
    >>> solver = pycryptosat.Solver()
    >>> solver.add_clause([1, -5, 4])
    >>> solver.add_clause([-1, 5, 3, 4])
-   >>> solver.add_clause([-3, -4]])
-   >>> pycryptosat.solve()
+   >>> solver.add_clause([-3, -4])
+   >>> solver.solve()
    (True, (None, True, False, False, True, True))
 
 This solution translates to: x\ :sub:`1` = x\ :sub:`4` = x\ :sub:`5` = True,
--- a/python/pycryptosat.cpp
+++ b/python/pycryptosat.cpp
@@ -33,10 +33,31 @@ THE SOFTWARE.
 #include <cryptominisat5/cryptominisat.h>
 using namespace CMSat;
 
+#define MODULE_NAME "pycryptosat"
+#define MODULE_DOC "CryptoSAT satisfiability solver."
+
+#if PY_MAJOR_VERSION >= 3
+#define IS_PY3K
+#endif
+
+#ifndef IS_PY3K
 #define IS_INT(x)  (PyInt_Check(x) || PyLong_Check(x))
+#else
+#define IS_INT(x)  PyLong_Check(x)
+#endif
 
 #if PY_MAJOR_VERSION == 2 && PY_MINOR_VERSION <= 5
 #define PyUnicode_FromString  PyString_FromString
+#endif
+
+// Should only be necessary on Python <2.6
+#ifndef Py_TYPE
+    #define Py_TYPE(ob) (((PyObject*)(ob))->ob_type)
+#endif
+
+#ifndef PyVarObject_HEAD_INIT
+    #define PyVarObject_HEAD_INIT(type, size) \
+        PyObject_HEAD_INIT(type) size,
 #endif
 
 typedef struct {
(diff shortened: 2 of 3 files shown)


msoos:
Wow, that's some pretty good work there, thank you!. I'm really impressed. Can I ask what you use the solver for? Any suggestions you may have? I'm really interested. Thanks again!

Mate

sliedes:
Mostly solving or trying to solve different combinatorial problems for fun. Though this I wrote for prototyping - I don't think Python is my final language of choice - simplifying and-inverter graphs (AIGs), which I use as building blocks to construct larger problems. I've done that using some ad hoc Python implementations for a while, until I recently discovered that Verilog is quite decent for specifying combinatorial AIGs in a high level language (there's a nice open source Verilog sythesis tool called yosys).

By the way, I noticed the CMakeLists still only targets Python 2.7, though Python 3 works if I set the cmake variables by hand. I wonder what should be done about that; should there be a switch to choose which Python to target (at least on my system both coexist)? Or try one first, if not found, try the other?

msoos:
yosys is cool, I met the author at CCC (33C3, last end of the year in Hamburg). Here is the video: https://media.ccc.de/v/33c3-7922-formal_verification_of_verilog_hdl_with_yosys-smtbmc If you have any questions about yosys, he is really helpful. Also, come to 34C3 and then we can meet :)

By the way, you probably know it, but libabc is kinda cool and may help you: https://bitbucket.org/alanmi/abc also, Armin Biere is amazing and his aiger stuff you are probably aware of. I am kinda interested what could be done -- sometimes, CNF can be converted into other formats suchh as AIG, simplifications carried out there, and some base lemmas (such as unit and binary clauses) can be imported back to the CNF. I do that with e.g. Gaussian elimination (though I do Gauss on-the-fly, too but that's another story)

As for CMake targeting -- maybe we could detect version somehow? Trying one first and building the other if it fails would be great thing, yes :)

sliedes:
Heh, I might well consider coming to 34C3. Actually sounds pretty interesting, and Hamburg is not that far away from Helsinki where I live.

Yes, I'm aware of abc and AIGER, and also occasionally use them to work on CNF. abc is nice, if unfortunately a bit unstable, and I find its source quite off-putting. If I had infinite time, maybe I'd look at replacing minisat (or actually I think several different versions of minisat...) inside abc with something more modern, and modify it to not abandon search in e.g. FRAIG after a small number of conflicts. As it stands now, I can for example load a single-output unsatisfiable .aig into it, and yet I think there's no command that reliably spends however much time it needs and simplifies it to the false literal. I think this is because abc is more geared towards hardware synthesis. Anyway, I'd really love a tool that takes an AIG which is not too tough for SAT solvers and reliably simplifies it so that at least inputs that don't have an effect become disconnected. I think that would be really useful for composing circuit-sat problems from smaller parts (or Verilog modules; I have a set of scripts to do that while optimizing each module with abc, but they're quite hackish).

Actually, I [sometimes wonder](http://cs.stackexchange.com/questions/71146/why-do-all-recent-sat-solvers-work-on-cnf-instead-of-circuit-sat) what happened to SAT solvers that work on AIGs.

CMake. I think a switch would be nice in any case. Probably any modern Debian or Ubuntu installation is going to have both Python 2 and Python 3 installed anyway (though not necessarily the necessary -dev packages).

msoos:
abc's source is rather... difficult to comprehend? :) I wish I could understand it, because it really needs to be fixed in many ways I think. Replacing the SAT solver is one of them. I thought Armin Biere's tools are the right thing for this. Simplification at the AIG level is interesting, but some (most?) of it can be done at the CNF level, and most of it is often not needed. What's great about SAT solvers is they (tend to) concentrate on the "hard" part of the system, and so only try to solve/search around there. AIG does not "know" (heuristically detect) where the interesting part is and hence applies its simplification to every part of the problem. If you are asking me, I think this is the reason why lots of people don't simplify their AIGs and inputs so much. They rely on SAT's heuristics (mostly VSIDS + clause learning-based variable activity bumping) to find the interesting part of the problem and simplify it using clause learning. Anyway, that's my guess. You also have to see that most people are horrendous when it comes to translating their problems to AIG/CNF. I mean, they blow it up by exponential factors etc. There needs to be a comprehensive book about this, probably... Somebody needs to write it in a way that is understandable to a much wider audience, that's my guess.

Looking forward to seeing you at 34C3 :)

PR #395, opened 2017-05-04
Update clauseallocator.cpp

cunnilingus123:
max(a,b) is only defined, when a and b have the same type. I made a type cast for that. The project is not compilable in mac (clang) else.


msoos:
Thanks! I fixed it in the same way... I didn't see your pull request :( But we think the same way, I like it! Do you have any other suggestions? I promise I'll merge the pull request :)

PR #396, opened 2017-05-14
cryptominisat5Config.cmake.in: fix exported library name

Mic92:
Otherwise the linker will search for `liblibcryptominisat5.so`


msoos:
Thanks, good catch!

Issue #367, opened 2016-09-01
Make it possible to compile python bindings without CMake.


arrowd:
Preprocessing done by CMake on `setup.py.in` is mostly unnecessary, but it causes some [inconveniences for packagers](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=211460#c1). Here is what we are currently doing to avoid that: https://svnweb.freebsd.org/ports/head/math/py-cryptominisat/files/patch-setup.py.in?view=markup

Could it be possible to convert `setup.py.in` into a plain `setup.py` ready to be consumed by PyPi?

msoos:
First of all, thanks for the heads up and the work! I am having trouble here, because this thing is supposed to work under 4 operating systems: Linux, Windows, MacOS, and FreeBSD. I can't possibly do that without some magic. CMake does a lot of the heavy-lifting and I love it for that. Is it an issue to keep that patch around for FreeBSD? I can add it to the source code if you feel like that makes sense. Unfortunately, I can't test it though, but I can try to make sure it at least applies cleanly, though I haven't changed that file for a while. Is there some kind of compromise we can do maybe? Thanks!

Mate

arrowd:
This is what i came out with: https://github.com/arrowd/cryptominisat/tree/python_without_cmake

With this i've managed to build python bindings as part of CMake build, as well as separate build against installed cmsat.

I'd like to see your comments on this.

msoos:
Thanks! Here are some issues:
* Unfortunately, cmake 3.1 is a bit too new. Though I think it might work with 2.8.8 as well?
* I'm afraid it will not work on Mac OS X.

However! We can have both a version of both setup.py and setup.py.in in there somehow, so it would work anyway. I'll think about how to best keep them both, so when cmake is not there, it would still work.

By the way, how is cryptominisat compiled without cmake? I'm wondering if removing cmake dependency  from the setup.py is useful, as cryptominisat itself would still need cmake :S

arrowd:
> Though I think it might work with 2.8.8 as well?

Alas, no. The `-E env` command first appeared in CMake 3.1.

> I'm afraid it will not work on Mac OS X.

Why do you think so?

> By the way, how is cryptominisat compiled without cmake?

Cmsat itself is compiled using CMake, of course, but we package python bindings separately from the mail package. Due to that, python bindings are compiled separately from cmsat and this is where CMake dependency kicks in. Having it only to pass some env vars is a bit of overkill.

[...]

PR #397, opened 2017-05-17
5.0.1 cmake dev

ysard:
Hi, i'm currently maintaining a project developped at the Inria/Irisa of Rennes - Bretagne Atlantique;
[this project](https://gitlab.irisa.fr/0000B8EG/Cadbiom/tree/pypi_packaging)
is developed in Python and is mainly based on pycryptosat.

To facilitate the maintenance work during the update of the dependencies I worked on the packaging of pycryptosat.

This code was made in February 2017 and is based on the numbered tag 5.0.1;
changes have been made on the master branch,
therefore I will work on the modifications to be made for a full merge.

Currently, In the modifications I am proposing, you will find these features:

- Support of Python2 and Python3 (depending of the compilation environment), as requested in #369;
- pycryptosat is still dependant of cmake to generate the setup.py, but it is now compiled in static mode
(the Python binding is no longer linked to the library);
this update allows packagers to upload your solver on PyPi, as requested in #367
So you will find these directives in the Cmake environment:
    - make python_interface: compilation of pycryptosat,
    - make python_interface_install: compilation and installation of pycryptosat in the current environment,
    - make python_interface_dist_package: make a distribution package for pypi
    - make python_interface_test: run unit tests

In the Python binding:
- Documentation of most of functions,
- add_clauses(): add clauses from an iterable (avoid to call multiple add_clause()),
- nb_vars(): a method to get the number of literals in the solver for debugging,
- is_satisfiable(): a method to quickly get the satisfiability of the current problem without obtaining any solution,
- msolve_selected(): a method used to find multiple solutions to a problem.
The solver is ran in a loop and each previous solution will be banned.
- some improvements in the use of the Python API.

Thank you for your work on the solver,
regards.


msoos:
Thanks! I'll review and integrate these in the weekend! Lots of great work, thanks!

ysard:
Thank you for your prompt response !
I have made the necessary modifications to a clean merge.

A decision should be made on the changes made in the ddbec17 commit.
In C, string literals are not of const type, but we use C ++ and these lines
raise the warning "deprecated conversion from string constant to 'char *'".

Changes in this commit seem to affect the readability of the code,
especially since we know what we are doing by using an array of char* for the `kwlist` variable.

The subject was discussed  [here](https://mail.python.org/pipermail/python-dev/2006-February/060850.html) with Guido van Rossum.

So, for the moment I use the following directive to hide these warnings that are a priori specific to gcc 4.3+:

`#pragma GCC diagnostic ignored "-Wwrite-strings"`

msoos:
Waiiiiit. We would build the library twice. Once as a ".so" and then as a python module. This doesn't seem to make sense. The two builds would never be the same (build flags, etc). And I don't see why we need to build it twice -- can't we just use the ".so" built by the system? It's a bit confusing. I thought all we need is a wrapper, not a complete re-build. Maintaining this would be hard and it's a duplication of efforts, I think, no?

ysard:
Hi, In fact if the library is not compiled during the installation process of the Python package (ie if sources are not provided with it), you will find these cons:

- users have to be redirected to Github and where they will have to manually install the library on the system or in a local folder,
- this prevents the inclusion of the solver and its package in an automatic installation workflow,
- this requires some additional waste of time or additional knowledge (like the use of git, cmake and make).

For large libraries requiring many system dependencies, such deployment process can be justified because it can be tricky to automate; but here, the compilation goes very well and is very easy with basic compilation flags.

With this setup.py users can install the solver and the Python binding with only one command:
`pip install pycryptosat`

And you can easily maintain your project with this couple of commands:
`python setup.py sdist upload`

That's why i made the system library independent from the Python packaging and this facilitates the use case discussed in #397.

regards.

msoos:
Thanks for the thorough explanation, I'm convinced :) Sorry, I'm just
acutely aware of what build complexities can lead to in terms of
maintenance overhead. Thanks again,  I'll merge it in the weekend :)

msoos:
Thanks, done! To be honest, it's a bit of a shame, that now we will have two builds, but it might be worth it.

I removed some of the flags you added to the python build. Ultimately, this should work with Windows (right?), which may not have gcc-type flags. Also, it might need to work with other compilers, which may have different types of flags. The stdc++11 will already be an issue for Windows, I'm sure :S The rest... optimisation should be default, which is OK.

msoos:
Great work, thanks again for all! Do you know how to put this up for python pip? I have never done that...

ysard:
Thank you for the merge and your time passed on it :)

I will try to work on a Windows/Linux compatible code (nmake, mingw32, gcc, clang).
Indeed, it might be worth it.

For the upload to pypi you can follow this [tutorial](http://peterdowns.com/posts/first-time-with-pypi.html).

Basically you will need to create an account on the 2 servers of pypi
([](https://pypi.python.org/pypi) & [](https://testpypi.python.org/pypi))

Then you will have to create a file `~/.pypirc` with your logins.

In fine this command will allow you to upload the package on the test server:
`python setup.py sdist upload -r pypitest`

(It appears that the registering step is no longer necessary)

NB: The test server is useful because any modification on the package
requires that you increment the version number for the upload on the public server of pypi
(they don't allow 2 different packages with the same version).

Issue #406, opened 2017-07-25
Remove dependency on vim (xxd)


horenmar:
For Windows, Vim is a a relatively obscure dependency and CMSat uses only a very small and simple part of it, and as far as I can tell, it is necessary to compile core of CMSat.
 
Other dependencies are either easy to get on windows (boost and zlib are both on vcpkg), or very optional (Python includes for pycmsat), so this would make cmsat easier to compile on Windows.

msoos:
Thanks, I merged it :) I think it's still better than vim :D

horenmar:
Ideally this would be done in CMake, but I don't know enough CMake. Otherwise the idea is that Python runtime is much more common, so while it just trades one dependency for another, the new dependency should be less of a burden when installing.

Overall I would keep the issue open for the future.

msoos:
I dont like hoarding issues :) It's fixed and done. If anybody has trouble,
 they'll open a new issue anyway, nobody bothers to read through old issues
:)

msoos:
Just as a feedback:
1) I like we moved away from this dependency :)
2) The xxd python file didn't have the right permissions (+x) set
3) The xxd python file generated a string that didn't have the ending NULL character and hence caused a classic memory overflow bug.

Overall, again, I'm happy :) So, the above is just feedback so you can avoid these issues in the future! Thankfully, most of these were caught by the test suite quite easily. Some were caught by others using the system -- many eyeballs... ;)

Cheers,

Mate

horenmar:
2 is fair, I work off Windows/WSL neither of which has much use for execute permission, so that slipped right past me.

However, either I have bugged local copy of `xxd`, or there was an error there previously, because after cloning current master (so it includes the python-xxd fix), I get this
```
/mnt/c/ubuntu/cryptominisat$ xxd -i cmsat_tablestructure.sql xxd-sql.cpp
/mnt/c/ubuntu/cryptominisat$ ./scripts/xxd-alike.py cmsat_tablestructure.sql xxd-alike-sql.cpp
/mnt/c/ubuntu/cryptominisat$ diff -w xxd-alike-sql.cpp xxd-sql.cpp
564c564
<   0x0a, 0x00
---
>   0x0a
566c566
< unsigned int cmsat_tablestructure_sql_len = 6746;
---
> unsigned int cmsat_tablestructure_sql_len = 6745;
```
Which means that previously the python script generated exactly the same output as using `xxd` and now it generates different output.

[...]

PR #407, opened 2017-07-25
Replace dependency on XXD with simple Python script

horenmar:
Python runtime is more common on Windows than Vim (on Linux its a wash), so I suggest trading dependency on vim for dependency on Python runtime. This PR implements required part of xxd in Python and is Python 2 and Python 3 compatible.

This could also likely be implemented in CMake, but not by me.

Related to #406


msoos:
To be fair, this is a lot better than vim :D Merged!

PR #409, opened 2017-07-29
Fix const placement in ReduceDB::get_total_time()

horenmar:
`const double` as a return type makes no sense and the method is const, so I moved it around to silence a warning from gcc.


msoos:
Wow, good catch!

PR #410, opened 2017-07-29
Properly delete CMSatPrivateData copy constructor (and assignment!)

horenmar:
This moves potential errors from runtime to compile time, and also
catches copy assignment errors.


msoos:
Thanks, great work! Forgot about the "= delete" :O

PR #411, opened 2017-07-29
Enable W4 when compiling with MSVC

horenmar:
Also fixes most of the resulting warnings. There are some 30 warnings
left, but they should be easy to fix with some mechanical work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,8 +216,14 @@ else()
     # buffers security check
     add_compile_options(/GS)
 
-    # not too many warnings
-    add_compile_options(/W1)
+    # Proper warning level
+    add_compile_options(/W4)
+    # Disable STL used in DLL-boundary warning
+    add_compile_options(/wd4251)
+    add_compile_options(/D_CRT_SECURE_NO_WARNINGS)
+
+    # Wall is MSVC's Weverything, so annoying unless used from the start
+    # and with judiciously used warning disables
     # add_compile_options(/Wall)
 
     # /Za = only ansi C98 & C++11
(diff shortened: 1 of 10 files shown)


horenmar:
The `Lit l` changes are there to remove variable shadowing warnings, as those scopes already had `Lit lit` previously declared. Same goes for the other small mechanical changes.

> /W4 may spit out WAAAAY too many warnings. I had at one time >100MB of warnings.... I am afraid. I will never be able to wade through that. Why do we need /W4? I am really afraid.

As mentioned, after the changes and using VS2017, only ~30 warnings are left and they are mostly mechanical in nature (and if solverconf.h was sanitized, there should be even less). The three biggest offenders were "stl in DLL exported function", which I decided to turn off project-wide, "constant expression in an if" caused by `release_assert(false, ...)` and ton of "assignment in an if" in the simple main, which got turned off for that file.

As to the reason it is needed, /W4 is needed for roughly the same reason that -Wall -Wextra is needed. It is a fairly reasonable set of warnings to tell you when you are doing something fishy.

msoos:
Ah, cool, makes sense :) Thanks! I'll merge tonight :) Thanks again!

msoos:
I get 4000+ lines of warnings in AppVeyor, please see

https://ci.appveyor.com/project/msoos/cryptominisat/build/1.0.396

This is really ugly, I want to be able to work with this, and warnings that are never resolved will not be useful, just clutter everything up. Do you know how to fix them? Can you please pull in a merge request that fixes it or pulls down the warning level please! Help would really be appreciated :) Thanks!

msoos:
I have created a new issue, here: https://github.com/msoos/cryptominisat/issues/419

horenmar:
Quick check of the log says that most of the warnings are reported over and over and over again, as they are caused by code in headers...

I'll throw up a PR to reduce them somewhat soon.

PR #416, opened 2017-08-09
Correct loop iterator name

curtisbright:
There was a typo in a loop iterator name which was introduced recently. (#411)


msoos:
Wow. What a catch. Thank you!

horenmar:
Welp, that was a dumb mistake.

capiman:
Could there be a testcase to catch this?

msoos:
Only if you make one :) Sorry, I'm short on time! Good luck!

Issue #419, opened 2017-08-15
Way too many warnings in MSVC build


msoos:
Will make /W1 unless it's resolved

horenmar:
After further investigations, there are about 200 warnings that are actually 2 warnings, but reported over and over and over and over again, because they occur in frequently included header files. I will open a PR fixing those soon.

Then there is `cnf-utils` situation, which has CMakefile that unconditionally adds `-Wall` to flags. Somehow that gets transformed into `/Wall` for MSVC, which is... not a good idea, as under MSVC it is an equivalent to Clang's `-Weverything` and includes things such as
```
warning C4514: 'XXX': unreferenced inline function has been removed
warning C4710: 'XXX': function not inlined
warning C4711: function 'XXX' selected for automatic inline expansion 
```
and other similarly noisy warnings.

Those need to be fixed separately, since they come from different repository.

msoos:
OK, I have merged the one for -Wall guarding, thanks! Let's see what happens now :)

msoos:
We are getting there :) Only about 2500 instead of 5000 lines of output! Thanks... soon it'll be quite manageable!

msoos:
I'm tuning down the MSVC warning levels. I don't see the reason why it is needed, as I am catching all issues in clang&gcc in Linux, and that should be sufficient. Please let me know if you must have W4 for some reason. I am reluctant to have (void)xyz; in the code to satisfy MSVC :)

horenmar:
Mostly I saw couple of easy improvements to CMSat that I wanted to contribute before moving on, so I don't really need it. However, it turns out that CMSat is outperformed by Minisat/Glucose for my problem, especially when I use knowledge of the original problem to hint polarities over individual variables.

msoos:
:) Cool! I will tune the W4 down. I am fully supportive of Windows builds and I am committed to make it work on Linux+Windows+OSX. This is not abandoning the Windows build at all. This is just making sure that the output in AppVeyor can be parsed by a human so it can be fixed -- in other words, it's precisely because I care about the Windows build and want to be able to maintain it.

I am *very* happy for any Windows build advice you may have that doesn't break Linux+OSX builds and keeps AppVeyor happy & its output parseable :)

Thanks again and keep the issues/merge requests rolling :)

PR #420, opened 2017-08-16
Reduce number of MSVC warnings

horenmar:
This fixes some 200 of the new warnings. After the ones originating in cnf-utils submodule are fixed, I intend to go through the remaining ones again and see how much is there left.

Related to #419 and #411


msoos:
I'm sorry, these are ultra-ugly, calling/dunno-what some variables just to make the compiler happy. I'm sorry, but I can't merge this with a good heart :( I'm sorry. I have merged the xorfinder.h, though, thanks!

horenmar:
It is casting them to void (ie doing nothing). It is actually common pattern to signal that you aren't using the variables intentionally. (In release build they are unused, as the asserts go away)

PR #424, opened 2017-08-26
typos

dennis714:


a1880:
While you are at it:
The plural of matrix is actual [matrices](https://www.google.de/search?q=matrixes+matrices) rather than matrixes.

msoos:
Both matrixes and matrices are correct. Please look at the dictionaries online or offline. Here is one:

http://www.dictionary.com/browse/matrix?s=t

Oxford dictionary will say the same. So will the American Heritage. Both are correct.

msoos:
Thanks for the typos fixes!! Really good stuff :) Merged, thanks!

a1880:
Well, I don't want to fight for matrices, but the Google hit count speaks a clear language: 40mio vs 400k

PR #458, opened 2018-02-18
Clarify performance impact of detailed statistics

saraedum:
The statement "leads to faster system" is a bit confusing. One immediately understands this as "no statistics ⇒ faster system" but it could also mean "no statistics. Too bad, they would lead to a faster system." I slightly stumbled upon this here https://github.com/conda-forge/staged-recipes/pull/5185.


msoos:
Thanks for the fix!

Issue #471, opened 2018-03-16
Add functionnalities to the Python interface


MPaulmier:
As said in #454 we've been working on the python interface for the past two months as part of an assignment. The goal was to add DRAT capabilities to the interface.

We're almost there and would like to know if this is something people would be interested in. Here is a summary of what we have done :

- added a `drat` argument to the `solve` function to query the creation DRAT file. This argument is a string which represents the name of the file to be generated (in the same way the command line application works);
- modified the `setup_solver` function to generate the entire python object and not only the SATSolver. Before that, this object only contained one field. Now it contains vectors that represents the list of [xor_]clauses. It looked more natural to generate the python object this way than generating each field in their own setup function;
- added the nb_clauses function. It only uses a counter that is incremented at each `add_[xor_]clause` (as suggested in #454);
- added a save function to save the state of the `Solver` object in a DIMACS file. This is needed if you want to verify your proof with DRAT-trim and have only generated the clauses through the interface;
- added a load function to load any DIMACS file at any moment in the Solver construction process;

Of course some other functions have been added as a side effect in the process.

This ended up being a bit more than just adding the DRAT generation to the interface as we realized we needed some other features.

What still needs to be done :

- add tests to the testsuite. ATM only the `nb_clauses` function is tested, which is... not much;
- we forked at 979d967 so their may be some changes that need to be done, however nothing has been done on this side to the interface so it should be pretty smooth;
- modify the README file to show usage of the new features;

Their's probably some work to do on the code as we're no experts in C++ let alone the implementation of Python bindings. We understand that this part of the project is not your primary focus and we'll have to work for our finals soon so it will probably take some time to clean up our repo and merge.

Improvements that could or will be done to our code :

- make the saving of the clauses non-mandatory. That's 3 vectors we added to the object (one for the regular clauses, the other two for the xor_clauses and their respective `rhs` variables). They may take a lot of memory and we don't always want to rebuild the DIMACS file from the object;
- we may have omitted some cases that are not compatible with the DRAT generation, if so, they need to be taken into consideration;
- some functions are really big and don't seem very efficient. We need to look into that;

As you may have realized, this is not a fast going project and we still have a lot of work to do on this. You can find our work in my repo here : https://github.com/MPaulmier/cryptominisat

[...]

PR #472, opened 2018-03-17
Add functionnalities to the python interface

MPaulmier:
Please see #471 for more details on this PR. This is an ongoing project and some parts still need to be discussed and reworked. 

- [ ] Address the memory issue caused by the added vectors
- [ ] Add tests to the test suite
- [ ] See if there are other problems we didn't think of

You'll see that we did a lot of experiments before finding a good way to generate the different files, it might look a bit messy in the commit messages.


MPaulmier:
Ok the changes to `cryptominisat.cpp` and `cryptominisat.h.in` shouldn't be in the repo. They're the remains of what we tried to do at the very beginning of this project. Should the responsible commit (30a6069) be reapplied without the changes to these files ? I don't know if that would work with the PR.

msoos:
Some comments:
Here you are changing the indentation. Please don't do that. Please add a commit to fix the indentation.
```
 -    PyObject *clause;
 +	PyObject* clause;
      if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &clause)) {
```


Here you have things indented to different indent even though they are in the sample place logically. Please never do this, **no project will ever accept** a pull request like that. Fix please.
```
PyObject* int_vector_to_list(const std::vector<int> &data) {
 +  PyObject* listObj = PyList_New(data.size());
 +	if (!listObj) PyErr_SetString(PyExc_SystemError, "unable to allocate memory for Python list/");
```
This is *incredibly* inefficient:

```
line.substr(0,1) != std::string("c")
```

What you want is:

```
line[0] != 'c'
```
The above one probably runs about 10x faster. Please be considerate about both memory usage and speed of execution.

MPaulmier:
These commits should fix indentation and code inefficiency.

MPaulmier:
About the memory issue, we're not so sure how to make it better. As you said our solution takes roughly 3x more memory than what we had previously. Taking the clauses directly from the `SATSolver` object isn't an option as far as we know since they will be altered in the process. That's why we went this route. Here are some thoughts we have on this : 

- making the recording of clauses non-mandatory would be a starting point. We would just have to set a flag in the setup function to ask for clause recording;
- we could try and find a better way to store them (other than two vectors of vectors of uint/lit) but this would still take more memory than the current implementation (roughly twice as much if we follow you implementation I would guess);
- we could simply get rid of it and assume the user would build their own file when generating the `Solver` object. In this case, what would be left of our work would be the drat file recovery, the `load` function for loading DIMACS files, the modification of the `setup_solver` function as well as the `nb_clauses` function. These functionnalities shouldn't penalize pycryptosat's performances;

msoos:
Please write the CNF to a temporary file instead of storing it in memory. No need for in-memory storage.

Also, why do we have a DIMACS parser included?

MPaulmier:
Ok this seems like a better idea in terms of memory. We started with that idea in mind and transitionned to the Python object for some reason I can't remember.
The DIMACS parser is required because DRAT-Trim takes a DIMACS file as an input. Since we're in the interface we only have the Python object and getting the drat file wouldn't make sense without this file.

msoos:
We still don't need a parser. We only need to write that DIMACS for DRAT to eat it. No need to parse it.

I still don't much get the point of this whole change request... Apart from providing something that a programmer can do in less than 1 minute (write to a file the clauses while adding them), it provides no effective value to the user, but actually creates a number of real pain points, including coding that doesn't care about user's memory, adding code bloat (DIMACS parser).

In the end, I will have to maintain this which won't be easy as this change provides no effective tests. There is no test that checks that DRAT actually eats the file output. No test that the file is actually DIMACS (header correct, etc), no test that the input clauses are the same as the ones put into DIMACS, etc. Also, XOR clauses cannot be eaten by DRAT, so that's another thing, how is that being handled? This looks like a hack :(

msoos:
I suggest you clean this up, give a rationale, add extensive tests, check with someone if they think it's OK and then create a new pull request! Thanks!

PR #478, opened 2018-04-05
Add max_time to Python wrapper

cxjdavin:


msoos:
Wow, this is really nice! Just one fix: the timeout is a double. You put it as a "long" in Python and "uint64_t" in C++. However, it's a double in CryptoMiniSat. Can you please change all of them to reflect that? I am afraid of overflow/underflow etc. Double and uint64_t can hold very different types of values.

You can fix the issue by simply pushing commits to the branch that you asked to merge. So no need to do anything here in GitHub -- if you push to that branch in your machine, it will automatically appear here and the pull request will be amended. So just add a commit to fix these :)

Also, can you please add a test to the python and the C++ codebase as well? No need to time out, but just that it actually works with a long timeout (doesn't give None/l_Undef for example), would be nice. I actually have tests for max_confl if you take a peek. Note that CMS is harder to develop precisely because I want to have tests. I know they are cumbersome but that's what makes the difference between academic junkware and systems that work in an industrial setting.

cxjdavin:
Yes, silly me. Of course max_time and time_limit should be "double". While at it, I also fixed "int" to "long" for confl_limit.

I've added test for invalid initialisation values for time_limit, but I don't see how your tests in the python/tests folder tests for confl_limit. TestSolve.setUp() doesn't actually set confl_limit, so it's not tested right?

msoos:
Thanks! Shame it's missing the tests, I'm adding them now :)

PR #483, opened 2018-04-13
fix USE_GAUS->USE_GAUSS

### LINKED ISSUE #482: USE_GAUS instead of USE_GAUSS in ifdef?
https://github.com/msoos/cryptominisat/commit/10387761634fdcedb81eab3ec8445b8680c79b57#diff-12e9d4787317f22b8230f6fe68594463R442

There is 
#ifdef USE_GAUS
instead of
#ifdef USE_GAUSS
???

cipherboy:
Fixes #482 if it is an issue. :) 

(We'll see if it breaks the CI -- local tests seem fine...)

Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>


msoos:
Ahhhh so that's why it never got triggered! Wow, good eye!!!

PR #490, opened 2018-05-16
Fix DESTDIR support for installing the Python module

embray:
CMake normally supports `DESTDIR` installation of files--this is used (typically by running something like `make install DESTDIR=some/tmp/dir`) to install packages into an alternate root, usually for packaging purposes.

However, when packaging cryptominisat the Python module is not installed into a supplied `DESTDIR`.  This could be worked around in principle by disabling CMake from installing the Python module, and supplying the necessary commands manually.  But better to just fix this directly.  This solution worked for me.


msoos:
Thanks, that's a cool fix!

PR #493, opened 2018-05-29
Add Fedora build instructions

cipherboy:
I ended up reorganizing the sections to place all dependency installation in the "Compiling in Linux" section. This then has an Ubuntu/Debian and Fedora/CentOS/RHEL section. Later sections point back to this section.

I also noticed that the Ubuntu instructions pull in `libzip-dev`, but shouldn't this be `zlib1g-dev`? `libzip-dev` is the dependency for parsing zip archives, whereas `zlib1g-dev` is for gzip streams; iirc CMS supports the latter. This is correct in the Fedora building instructions.

Thanks!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -70,12 +70,28 @@ docker run --rm -i -p 80:80 cmsweb
 Compiling in Linux
 -----
 
-To build and install, issue:
+To build and install on Ubuntu, issue:
 
 ```
 sudo apt-get install build-essential cmake
 # not required but very useful
-sudo apt-get install libzip-dev libboost-program-options-dev libm4ri-dev libsqlite3-dev
+sudo apt-get install zlib1g-dev libboost-program-options-dev libm4ri-dev libsqlite3-dev
+tar xzvf cryptominisat-version.tar.gz
+cd cryptominisat-version
+mkdir build && cd build
+cmake ..
+make
+sudo make install
+sudo ldconfig
+```
+
+
+To build and install on Fedora, issue:
+
+```
+sudo dnf install cmake valgrind valgrind-devel gcc gcc-c++ automake make autoconf help2man
+# not required but very useful
+sudo dnf install zlib-devel boost boost-devel boost-program-options m4ri-devel sqlite-devel
 tar xzvf cryptominisat-version.tar.gz
 cd cryptominisat-version
 mkdir build && cd build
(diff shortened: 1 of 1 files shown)


msoos:
Sorry, but the "install required dependencies as per XXX" sounds great but in practice, people don't give a shit -- they want to copy-paste the whole thing. Also, we don't need the "# if YY is required then the below" because most people have absolutely no idea what "python" actually means. They are using python without even thinking what it is. Basically, you must assume that people are (1) lazy and/or (2) have no time to click around. If you tell them:

"If X please Y and by they way also Z if you want K"

what you expect them to do is:

"Oh, I want X so I need Y and no need for K so no Z"

what they actually do:

"Oh fuck this shit, let me just use MiniSat"

cipherboy:
Understood; I just wish installing packages was a little easier between Linux flavors. :) Combination of different package managers and different package naming conventions leaves a bit to be desired. 

I'm (personally) hesitant to include both Ubuntu and Fedora versions for each of the major install blocks (Compiling / Python / Gaussian E. / Test Suite), as some of those should also work on macOS / Windows...

What do you think about compressing them to a single line having only one set of build instructions? Then `cmake` flags can be explained (`-DUSE_GAUSS=ON`, `-DENABLE_TESTING=ON`, etc.) in one location for them to adjust the build options. The later sections (on Gauss/Python. etc) can be simplified to be about usage instead of compiling+usage. 

E.g.,:

# Compiling in Linux

To install prerequisite dependencies on Ubuntu:
```
sudo apt-get install build-essential cmake libzip-dev libboost-program-options-dev libm4ri-dev libsqlite3-dev python3-setuptools python3-dev graphviz python3-numpy
```

To install prerequisite dependencies on Fedora:
```
sudo dnf install cmake valgrind valgrind-devel gcc gcc-c++ automake make autoconf help2man zlib-devel boost boost-devel boost-program-options m4ri-devel sqlite-devel python3-devel python3-pip python3-setuptools python3-numpy graphviz
```

To build and install, issue:
```
tar xzvf cryptominisat-version.tar.gz
cd cryptominisat-version
mkdir build && cd build
cmake ..
make
sudo make install
sudo ldconfig
```

## CMake Arguments

The following arguments to cmake configure the generated build artifacts. To use, specify options prior to running make in a clean subdirectory: `cmake <options> ..`

- `-DUSE_GAUSS=<ON/OFF>` -- whether or not to build with Gaussian Elimination support
- `-DENABLE_TESTING=<ON/OFF>` -- whether or not to build with test suite support
- `-DNOVALGRIND=<ON/OFF>` -- whether or not to build with extended memory checking support
- `-DONLY_SIMPLE=ON` -- whether or not to build only the simple binary and library interface

... etc 


Thoughts?

README.markdown
```diff
@@ -458,21 +465,15 @@ cd ../cryptominisat/scripts/fuzz/
...
-sudo apt-get install graphviz
-sudo apt-get install python3-pip python3-setuptools python3-dev
-sudo apt-get install python3-numpy
 sudo pip3 install --upgrade pip
 sudo pip3 install lit
 sudo pip3 install scikit-learn pandas scipy
 git clone https://github.com/msoos/cryptominisat.git
 cd cryptominisat
 git submodule update --init
+git checkout clauseID
```
cipherboy: Do you want me to split this change out into a separate commit? In the current set of docs, the `git checkout` comes before the `git clone`, so it doesn't have the desired effect. Also, the `clauseID` branch seems stale; is this step still needed? This branch is 35 commits ahead, 9223 commits behind master.

msoos:
Ah, I really like this part:

---
CMake Arguments
The following arguments to cmake configure the generated build artifacts. To use, specify options prior to running make in a clean subdirectory: cmake <options> ..

-DUSE_GAUSS=<ON/OFF> -- whether or not to build with Gaussian Elimination support
-DENABLE_TESTING=<ON/OFF> -- whether or not to build with test suite support
-DNOVALGRIND=<ON/OFF> -- whether or not to build with extended memory checking support
-DONLY_SIMPLE=ON -- whether or not to build only the simple binary and library interface
---

So please let's just do that and please let's add a "how to build on Fedora for normal use" (i.e. just a base install on Fedora). That's all. I don't want to add all the different ways to build both for Fedora and Debian or it will go crazy :) So can we just reduce the changeset to adding these two subsections? Thanks!!

cipherboy:
Done :)

msoos:
I have now added the CMake parts to the readme -- unfortunately I cannot support all distros, e.g. Fedora, Debian and Arch. I personally use Arch actually so the README was not written for me, it was written in a way that most people can adjust it to their own distros. So I consider this done. Please let me know what you think. I think the readme is super-long already... :S

Commit was:

755b69db4b888537fdb2930e7fb50313ed09b7fd

msoos:
Thanks again for this, I think what came out of it -- a fully new paragraph -- was really worth it. I know it sometimes seems like not everything is taken on board but I must balance the length of the readme with  the value that it adds. I could have re-written the whole thing for Fedora, that was the only other option, as I cannot possibly add Fedora + Arch + Debian to it, or it will become too long :(

PR #494, opened 2018-05-31
Fix man page install location

cipherboy:
Under 5.6.0, cmake configures the destination for the man pages to be `$PREFIX/man/`; however, man pages usually reside in `$PREFIX/share/man`. See FILES in `man man` ([here](http://man.he.net/?topic=man&section=all)).


msoos:
Ah, thanks, great catch!!

Issue #473, opened 2018-03-18
License


saraedum:
Trying to package this for Debian (#456), I stumbled upon some license question. The `LICENSE.txt` file says that cryptominisat is licensed under MIT (expat).

* [x] There is some code that is licensed under GPL-2 (without the option to upgrade to GPL-3). @msoos is the copyright holder of all this code, so maybe it would be an option to change the copyright headers to use the MIT license instead to make this consistent?
* [x] What is the license of `splncs.bst` and `ieee.cls`?
* [x] Am I right to assume that everything that has no copyright header is MIT licensed (I guess, I can imply that from the `LICENSE.txt` file.)
* [x] `cmake/GetGitRevisionDescription.cmake.in` is BSL-1.0 licensed but as I understand that's fine; cryptominisat can redistribute it under an `MIT` license without any explicit notice.

saraedum:
@msoos: Sorry for this issue. I know that working on licensing is not really productive but Debian is quite picky about these things.

msoos:
It is productive actually (I have worked in industry, I see the value), and I appreciate your work :) Also, please be aware that a MUCH faster version is on the way, so if you could package this, say, the 5th of March, that'd be the best -- the SAT competition deadline is the 1st so of course I shall have a fast solver then, which will get released.

For the individual pieces:
* splncs.bst -- just fixed this one. It now states the license ("This generated file can be redistributed and/or modified under the terms....")
* the only things missing headers should be trivial shell scripts. I am checking for headers for all C and H files and now for the fuzzer and AWS too.
* The python files are under GPLv2, including both the fuzzer and the AWS
* The AWS stuff now is under explicit GPLv2.
* Note that the actual library, headers, etc so stuff that is not for me (e.g. fuzzing with python, performance analysis with AWS) is all MIT and should not be in any other way.

Please let me know what is missing headers! I think all should have them... Should I put them on trivial shell scripts as well? They won't get installed...

As for my rationale:
* Everything needed to build both the library and the binary are MIT. Hence everything that gets installed is MIT. There is absolutely no worry for anyone that their code may be "infected" by GPL.
* All things to do with the python header and package are MIT
* All tests (whether python, C++, or other) are MIT
* Prediction scripts, fuzzing, and AWS are all GPLv2 because I want these things to be in the open. They are stand-alone and need no compilation. They should never be installed or ran on the user's machine, they are for development purposes only.

Please let me know what you think.

saraedum:
[...]

PR #495, opened 2018-05-31
Clarify license in pycryptosat

cipherboy:
In `python/LICENSE`, it states that CryptoMiniSat is under the LGPLv2. However, in [earlier](https://github.com/msoos/cryptominisat/issues/473) discussions, you've clarified that it is actually MIT. This updates the LICENSE documentation in `pycryptosat` to reflect that.


msoos:
Thanks!! Good catch!

PR #499, opened 2018-07-11
Fix DESTDIR support for the Python package again

embray:
Fixes regression against #490 introduced by 63a629f5e363a467abff1fa29becf42fe55b79f9 and b3cd73f369d4ea14f12824f0f3c4a06a9aa0e1b1 `CMAKE_INSTALL_PREFIX` is *not* the same thing as `DESTDIR`, and `--prefix` is not the same thing as `--root` (though both can coexist).

The installation path ends up being `<root><prefix>/<libdir>/<pythonX.Y>/site-packages/`.  By default root is normally `/` but the point of `DESTDIR` is to change that to an alternate path (e.g. for packaging).

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -61,7 +61,7 @@ add_custom_target(python_interface
                   ALL DEPENDS ${OUTPUT}/timestamp)
 
 install(CODE "execute_process(
-    COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} install --prefix=${CMAKE_INSTALL_PREFIX} --record files.txt
+    COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} install --root=\$ENV{DESTDIR} --prefix=${CMAKE_INSTALL_PREFIX} --record files.txt
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})"
 )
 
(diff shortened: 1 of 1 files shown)


cipherboy:
@msoos For reference, the Fedora package ([spec file here](https://src.fedoraproject.org/rpms/cryptominisat/blob/master/f/cryptominisat.spec#_75-76)) uses the contents of this PR to package CMS.

msoos:
Ah, that's good feedback y thank you! Okay, I'll try my best to fix the
surrounding issues and pull this ASAP :) Thank you both!

embray:
Supporting `CMAKE_INSTALL_PREFIX` properly definitely makes sense, and is probably something I should have included in my original fix.    Though `--root` and `--prefix` are really both necessary for packaging purposes.

embray:
I think I see what the problem might have been.  distutils has an annoyance that if `--root=` is passed an empty value, it takes the current directory as the root and strips the leading `/` off existing paths.  I didn't catch this because my use case was packaging with `DESTDIR` and I only tested with it, not without it.  Without `DESTDIR` it did not install properly.  That's my bad.

embray:
Confirmed that that was likely the issue.

msoos:
Thank you so much! Yeah, all checks have passed so I guess this should fix it for good. Thank you again!

PR #503, opened 2018-07-21
Update README to reflect zlib dependency

cipherboy:
Dropped when #493 was closed were changes to the README for Ubuntu. Your (admittedly old) Debian [control](https://github.com/msoos/cryptominisat/blob/bf3ec2968e488e5d4c31da08d27e5e2dda4c3d1d/debian/control#L5) listed zlib as a dependency as does the Fedora [spec](https://src.fedoraproject.org/rpms/cryptominisat/blob/master/f/cryptominisat.spec#_21). 

(Other than that, I'm fine with dropping the Fedora-specific changes to the README).

Thanks!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -75,7 +75,7 @@ To build and install, issue:
 ```
 sudo apt-get install build-essential cmake
 # not required but very useful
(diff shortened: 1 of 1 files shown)


msoos:
Sorry for the late merge! Thanks so much!

cipherboy:
Thanks @msoos! :)

Issue #511, opened 2018-08-21
Solver API for CNF output


janmazak:
I would like to know if there is a public API for getting the CNF
formula from CMSat::Solver (in DIMACS format). It is easy to add
clauses to a solver, so that is probably the simplest way to construct the
formula, but I do not know how to get it into a file (or, preferably, into a
string or sstream, which would be even better in use cases where many "small" formulas are to be processed). Perhaps also with an option to
get a simplified formula, but I am not sure what simplify() does --
the documentation does not say much (or have I overlooked it?) and
I do not know enough details on how SAT solvers work to guess what simplify() does.
(Note: the simplification available from command line via -p option is definitely useful, saves about 30% time on subsequent solving by plingeling, at least for the few instances I tried.)

There is only open_file_and_dump_irred_clauses(), but the output is
not exactly DIMACS and I am not sure what it actually is, and it only
works after simplify() is called.

And there is clausedumper.h, but it does not seem to be "public" API.

msoos:
You should be able to read in a file like:

```
void readInAFile(SATSolver* solver2, const string& filename)
{
    FILE * in = fopen(filename.c_str(), "rb");
    DimacsParser<StreamBuffer<FILE*, FN> > parser(solver, NULL, 2);
 
    if (in == NULL) {
        std::cerr
        << "ERROR! Could not open file '"
        << filename
        << "' for reading: " << strerror(errno) << endl;

        std::exit(1);
    }

    if (!parser.parse_DIMACS(in, false)) {
        exit(-1);
    }

    conf.independent_vars.swap(parser.independent_vars);

    #ifndef USE_ZLIB
        fclose(in);
    #else
        gzclose(in);
    #endif
}
```

If you include `dimacsparser.h`. It should be able to read in a string, too, if you use "CH" instead of FN and pass in a string. Check out stringbuffer.h and dimacsparser.h.

As for dumping a simplified file, it actually should now dump it correctly after simplification. Check it out, and let me know if it's wrong somehow. It should not be as far as I understand, there are even some tests for it I believe :) Let me know if the above helped enough or if you need more. Also, if it actually helped, may I ask you to write up a little how-to so I can put that into the README? I think others might benefit from your work :) Let me know if you have bumped into anything! Good luck!

janmazak:
[...]

PR #514, opened 2018-08-23
Allow dump into streams

janmazak:
Related to issue #511.

It compiles, but I have not tested it since I am not sure how to verify if the output is correct (refactoring might have broken existing dump into file); I hope there are some automated tests to check it :)

### DIFF
--- a/src/clausedumper.cpp
+++ b/src/clausedumper.cpp
@@ -28,41 +28,46 @@ THE SOFTWARE.
 
 using namespace CMSat;
 
-void ClauseDumper::write_unsat_file()
+void ClauseDumper::write_unsat(std::ostream *out)
 {
-    *outfile
+    *out
     << "p cnf 0 1\n"
     << "0\n";
-}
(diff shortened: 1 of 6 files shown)


msoos:
Sounds cool :) I need a small change, though. We cannot break the interface to the outside. So we'd need to create a new function, e.g. `dump_irred_clauses_ostream` for that, and keep the old one around (which would simply open the file and then use `dump_irred_clauses_ostream`). Similarly for `dump_red_clauses`. Would you be able to do that for me? Thanks!

janmazak:
As far as I understand, I have only added new methods to the public interface and kept the existing ones. (At `ClauseDumper`, I have changed parameters to constant strings, but that should not break anything?)

Could you please give me some more details on what to change?

msoos:
Oh, damn, I didn't read through carefully enough :O Sorry! I'll merge
tomorrow and fix the header too :) Thanks again!

PR #515, opened 2018-08-24
[WIP][PoC][testing] Python array.array interface

mbargull:
\[ DISCLAIMER: This is no production-ready code, just for testing and unfinished, plus I know next to nothing about how to efficiently work with Python C-API :wink: \]

This is in reference to
- https://github.com/msoos/cryptominisat/issues/513, and
- https://github.com/conda/conda/pull/7676

I put together some non-polished interface to be able to pass clauses to `Solver.add_clauses` as a Python `array.array("i")` object. This is so we can compare the time taken on processing the clauses as a simple stream of integers (separated by zeros) to processing them as iterables of iterables, e.g., list of tuples, which the current interface expects.
Since the time taken to scan an integer array is practically zero (if we need to see all values anyways, that is), the remaining time of adding clauses can be assumed to be spent outside of the Python interface, i.e., be admeasured to the core library.

I'll follow up with some measurement results..

PR #517, opened 2018-08-25
python: add_clauses: add max_var=0 kwarg

mbargull:
Adds `max_var=0` keyword argument to `add_clauses` in the Python interface.
This lets one specify the maximum variable number that is used in the given clauses and in turn enables speedups in case a high number of new variables is added.
xrefs: https://github.com/msoos/cryptominisat/pull/515#issuecomment-415856002, https://github.com/msoos/cryptominisat/pull/515#issuecomment-415901769

### DIFF
--- a/python/src/pycryptosat.cpp
+++ b/python/src/pycryptosat.cpp
@@ -348,10 +348,14 @@ Add iterable of clauses to the solver.\n\
 
 static PyObject* add_clauses(Solver *self, PyObject *args, PyObject *kwds)
 {
-    static char* kwlist[] = {"clauses", NULL};
+    static char* kwlist[] = {"clauses", "max_var", NULL};
     PyObject *clauses;
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &clauses)) {
+    long int max_var = 0;
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|l", kwlist, &clauses, &max_var)) {
         return NULL;
+    }
+    if (max_var >= (long int)self->cmsat->nVars()) {
+        self->cmsat->new_vars(max_var-(long int)self->cmsat->nVars()+1);
     }
 
     PyObject *iterator = PyObject_GetIter(clauses);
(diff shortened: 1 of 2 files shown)


python/src/pycryptosat.cpp
```diff
@@ -348,11 +348,15 @@ Add iterable of clauses to the solver.\n\
...
 {
-    static char* kwlist[] = {"clauses", NULL};
+    static char* kwlist[] = {"clauses", "max_var", NULL};
     PyObject *clauses;
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O", kwlist, &clauses)) {
+    long int max_var = 0;
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|l", kwlist, &clauses, &max_var)) {
         return NULL;
     }
+    if (max_var >= (long int)self->cmsat->nVars()) {
```
mbargull: I simply copied that logic from https://github.com/msoos/cryptominisat/commit/3a2b9a60e02e061df322a4664ce119f8c449bd75#diff-437bd5e874df2ba9f9ced39b13eee9abR185.
I'm a little unsure about the offset, though, as I'm too unfamiliar with CMS inner workings, e.g., in this case when zero- or one-based variable indices are used.
If we have `max_var == 0`, this will add a new variable, which seems odd. Is there always a base variable, i.e., is `self->cmsat->nVars() >= 1` always true?

python/src/pycryptosat.cpp
```diff
@@ -348,11 +348,15 @@ Add iterable of clauses to the solver.\n\
...
+        self->cmsat->new_vars(max_var-(long int)self->cmsat->nVars()+1);
```
msoos: This +1 is not needed here. The max_var is from the outside, hence it's e.g. 1 when there is 1 variable.

python/src/pycryptosat.cpp
msoos: This needs to be changed to `>` instead of `>=`. Here, the max_var is the actual variable number, not zero-starting.

mbargull: I see, so I guess in the part I copied from, the offset is different because of the `convert_lit_to_sign_and_var` call at https://github.com/msoos/cryptominisat/commit/3a2b9a60e02e061df322a4664ce119f8c449bd75#diff-437bd5e874df2ba9f9ced39b13eee9abL173.

mbargull:
Fixed the offset and added `max_var=m` to the performance testing script. So I think this is good to go now!

msoos:
Thanks, that's great :) And sorry for the slow response!

PR #518, opened 2018-08-25
Solver::addClause: use add_clause_int_tmp_cl

mbargull:
This is a follow-up to https://github.com/msoos/cryptominisat/pull/515#issuecomment-415870385 ff. and https://github.com/msoos/cryptominisat/commit/3baaafe678057a043f93569505da72071ea79a52.

We can avoid copying the input clause not only in `add_clause_int` but also in `addClause`, which gives another speed up.

### DIFF
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -638,6 +638,12 @@ bool Solver::addClauseHelper(vector<Lit>& ps)
 
 bool Solver::addClause(const vector<Lit>& lits, bool red)
 {
+    vector<Lit> ps = lits;
+    return Solver::addClauseInt(ps, red);
+}
+
+bool Solver::addClauseInt(vector<Lit>& ps, bool red)
+{
     if (conf.perform_occur_based_simp && occsimplifier->getAnythingHasBeenBlocked()) {
         std::cerr
         << "ERROR: Cannot add new clauses to the system if blocking was"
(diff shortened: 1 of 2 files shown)


mbargull:
Well, I think I'm missing something obvious here, since this change made a ton of tests fail.
I mean in https://github.com/msoos/cryptominisat/pull/515#issuecomment-415908564 you said that " `addClause` can recursively call itself", which would make this change problematic of course. But I think I'm currently too oblivious to where/how that recursion takes place.

mbargull:
So I reverted back to using that slightly ugly `addClauseInt` indirection which avoids recursion conflicts.
Running `scripts/performance/addclause.py`, we get
```
$ # master at https://github.com/msoos/cryptominisat/commit/e4282a5eee7c891db8aad364fe87e51b7bf84f37
$ python ./scripts/performance/addclause.py
setup CMS     list  0.539 0.540 0.565 0.571 0.578
setup Pycosat list  0.242 0.257 0.260 0.262 0.267
solve CMS     list  0.102 0.106 0.106 0.107 0.110
solve Pycosat list  1.583 1.587 1.589 1.590 1.592

$ # at https://github.com/msoos/cryptominisat/pull/518/commits/f05cac2da7d6941712c6a67d668a6da81700e9f2
$ python ./scripts/performance/addclause.py
setup CMS     list  0.490 0.493 0.511 0.516 0.521
setup Pycosat list  0.255 0.258 0.262 0.264 0.269
solve CMS     list  0.102 0.106 0.107 0.107 0.110
solve Pycosat list  1.587 1.587 1.588 1.592 1.604
```
so this would reduce the run time by `~10 %`.

If you have an equivalent solution that would avoid adding that `addClauseInt`, we can gladly use that instead :wink:

msoos:
Perfect, thanks!

PR #519, opened 2018-08-27
python: add_clauses: support array.array as input

mbargull:
Adds support to the `add_clauses` method of the Python interface for passing over clauses as a flat `array.array` of integers. Each clause has to be terminated by a zero, such that the integer array represents a stream of zero separated clauses.

In a subsequent step, this could be improved upon to support not only `array.array` instances but additionally support the [Python C-API buffer protocol](https://docs.python.org/3/c-api/buffer.html). If anyone with more experience around the Python C-API wants to volunteer for it, this would be highly appreciated!

### DIFF
--- a/python/src/pycryptosat.cpp
+++ b/python/src/pycryptosat.cpp
@@ -336,13 +336,143 @@ static PyObject* add_clause(Solver *self, PyObject *args, PyObject *kwds)
     return Py_None;
 }
 
+template <typename T>
+static int _add_clauses_from_array(Solver *self, size_t array_length, const T *array)
+{
+    if (array_length == 0) {
+        return 1;
+    }
+    if (array[array_length - 1] != 0) {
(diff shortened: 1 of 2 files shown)


mbargull:
```
$ python ./scripts/performance/addclause.py 
setup CMS     array 0.250 0.251 0.251 0.254 0.260
setup CMS     list  0.418 0.427 0.427 0.439 0.450
setup Pycosat array 0.178 0.179 0.180 0.180 0.182
setup Pycosat list  0.256 0.259 0.259 0.262 0.263
solve CMS     array 0.096 0.096 0.096 0.097 0.097
solve CMS     list  0.097 0.105 0.108 0.109 0.109
solve Pycosat array 1.585 1.586 1.588 1.588 1.590
solve Pycosat list  1.584 1.587 1.593 1.595 1.601
```

msoos:
Thanks, it's cool :) I'm going to add some tests now :) Merging.

PR #520, opened 2018-08-28
python: add_clauses_array: check itemsize

mbargull:
I was thinking it could make sense to use `array.itemsize` instead of `array.typecode` to decide which type the array is, in case different compilers (with `LP64` vs `LLP64`) were used to compile the interpreter and the extension. (The `typecode` is still checked to make sure we get an array of signed integers.)
It could very well be that this is not important. But it also doesn't hurt to do this :).

PR #521, opened 2018-08-28
python: tests: ensure str typecode for PY2

mbargull:
Fixes https://github.com/msoos/cryptominisat/pull/520#issuecomment-416742392.
Explanation: `from __future__ import unicode_literals` is used for forward compatibility to Python 3 (which is a good thing :+1:) and causes all string literals in the module to be of type `unicode` instead of `str`. For Python 3 `str` is equal to `unicode`. For Python 2 `str` is `bytes`, though, and `array.array` expects the first argument to be of type `str`, i.e., `bytes` for PY2.
Through a simple wrapper with `str(typecode)`, we can ensure that the argument becomes of type `bytes`.

### DIFF
--- a/python/tests/test_pycryptosat.py
+++ b/python/tests/test_pycryptosat.py
@@ -22,13 +22,17 @@
 
 from __future__ import unicode_literals
 from __future__ import print_function
+from array import array as _array
 import sys
 import unittest
 
 
 import pycryptosat
-import array
 from pycryptosat import Solver
+
+
+def array(typecode, initializer=()):
+    return _array(str(typecode), initializer)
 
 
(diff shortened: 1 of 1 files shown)


msoos:
Wow, nice investigation, I had no idea what was going on, that error thrown by Python was hard for me to grasp. Thank you! Merging :)

mbargull:
No problem, no real investigation needed :wink:, I just have encountered these things much more times than I've liked already (Damn you Python 2, be gone already!).

mbargull:
Some tests still failed because... \*facepalm\*... I also forgot to consider that the `typecode` is not `unicode` for PY2. Fixed that in gh-522.

PR #539, opened 2019-02-22
Support Paths with Spaces

dMaggot:
The way help2man generates the manpages, you need to pass the location
of the binary with escaped quotes.


msoos:
Wow, thank you so much!

Mate

PR #540, opened 2019-02-27
[WIP] Per-solve timeouts in the Python API

cipherboy:
Not sure if you're interested in this. 

I'd like a way to set timeouts for individual solves (versus all solves) -- the first solve takes the most time, and subsequent solves take less. (I'm also not particularly interested in finding *everything* as that's too big of a space. Just until it starts taking too long for some arbitrary metric...)

I figured I'd copy the semantics of the `Solver(...)` constructor and apply it to `Solver.solve(...)` in case others wish to add per-solve conflict limits, verbosity, or ~change the number of threads.~ **edit:** it appears that changing the number of threads can't be done after adding clauses.

After each solve, the `Solver`-level defaults are restored.

Since these are keyword arguments, these shouldn't break any existing program. 

(WIP since I'd like to add a test case).

Thoughts? Thanks!

### DIFF
--- a/python/src/pycryptosat.cpp.in
+++ b/python/src/pycryptosat.cpp.in
@@ -73,6 +73,10 @@ typedef struct {
     /* Type-specific fields go here. */
     SATSolver* cmsat;
     std::vector<Lit> tmp_cl_lits;
+
+    int verbose;
+    double time_limit;
+    long confl_limit;
 } Solver;
 
 static const char solver_create_docstring[] = \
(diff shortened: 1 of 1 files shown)


cipherboy:
Ah: You can't call `set_num_threads` per-run; only at the beginning.

msoos:
Ah, that's great! I didn't spend enough time on this, sorry! Thanks again :)

cipherboy:
Cool, thanks! I might have more changes in the future, we'll see. :)

msoos:
Thank *you*!

PR #542, opened 2019-03-05
Fix arguments in Python docstrings

cipherboy:
This changes all references to `arg[n]` to the specified `kwarg` name, moves
types to be with their parameter descriptions, and adds a missing
`max_vars` parameter to `add_clauses(...)`.

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`

PR #543, opened 2019-03-05
Add use max_var in add_clauses(...) test case

cipherboy:
The version of `pycryptosat` packaged by Fedora 29 currently is 5.6.4, which predates some of the recent improvements to the Python API. Not sure if you want it, but this ensures we at least attempt to test `max_vars` as part of the existing test suite. 

(`add_clauses` is called in other places without kwargs, so coverage should only improve... slightly :).

### DIFF
--- a/python/tests/test_pycryptosat.py
+++ b/python/tests/test_pycryptosat.py
@@ -204,9 +204,10 @@ class TestSolve(unittest.TestCase):
         self.assertTrue(check_solution(clauses1, solution))
 
     def test_add_clauses(self):
-        self.solver.add_clauses([[1], [-1]])
+        self.solver.add_clauses(clauses=[[1], [-1]], max_var=100)
         res, solution = self.solver.solve()
         self.assertEqual(res, False)
+        self.assertEqual(self.solver.nb_vars(), 100)
 
     def test_add_clauses_wrong_zero(self):
         self.assertRaises(TypeError, self.solver.add_clause, [[1, 0], [-1]])


cipherboy:
I noticed it wasn't tested after trying to figure out why it didn't work on my installed version of `pycryptosat`, figured I might as well add a test to make sure it does what I expect. :)

cipherboy:
@msoos What's the secret to the coverage test? The rest of the failures appear to be Ubuntu mirrors acting up (`E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/trusty-security/main/binary-i386/Packages.gz  Hash Sum mismatch`), but twice now I've failed coverage...

msoos:
I am VERY concerned about this "max_var" -- this MUST be removed. I believe you added it? It's very wrong. Please remove ASAP.

Thank you,

Mate

cipherboy:
@msoos `max_var` was added earlier: [commit](https://github.com/msoos/cryptominisat/commit/068965856241f8ccef8f6572ebb19856ccc17975), [PR](https://github.com/msoos/cryptominisat/pull/517). 

The last PR documented its existence, this PR tests it.

msoos:
I just removed it.

msoos:
OK, closing this because this max_var is very concerning to me.

cipherboy:
I think it was meant as a CNF size hint. The Python API doesn't otherwise expose `cmsat->new_vars(...)`. You end up calling it [during parsing](https://github.com/msoos/cryptominisat/blob/ffd757bf9470b1c471ab00a116b2f213f7ac9273/python/src/pycryptosat.cpp.in#L382-L386), so setting it ahead of time (via `add_clause`) buckets most of the cost of allocation up front vs. incrementally.

I think this is most helpful when you're building an entire CNF separately and passing it into CMS all at once. Clauses will likely go from low numbered vars -> high numbered vars in this case, so you're ending up resizing your model frequently vs. all at once in the beginning. I'm guessing this ends up being less expensive based on other discussions in that PR?

My 2c.

msoos:
Yeap, you are absolutely right! And exposing it would make some sense -- but people would most likely misuse it :(  We could experiment with it, but I am afraid. Maybe a separate function would make more sense -- that way, the experts could play with it. But putting into `add_clause` may make the average/newbie user use it, and that would lead to confusion and pain for them. Many people don't have the time to dig into all the docs, etc. so they might think it's some hint that doesn't have any drawbacks -- but it does, because if you pass, e.g. 100000 there, then it will eat a ton of memory and will take quite some time. But I see your point -- just let's try to make it harder to misuse :)

cipherboy:
So keep in mind this is a keyword argument (well, all of the Python API is keyword argument based), not a C/C++ function call.

If you're writing this API in Python, the function signature would look more like this:

```python
def add_clauses(clauses, max_var=0):
    pass
```

So most people will call it like this:

```python
solver.add_clauses(<my data>)
```

and won't realize that the `max_var` parameter is present/exists. The'd have to do one of these:

```python
solver.add_clauses(<my data>, <my num vars>)
solver.add_clauses(clauses=<my data>, max_var=<my num vars>)
```

to use it. Until #542, it wasn't exposed in the comments either, so `help(solver.add_clauses)` wouldn't inform you about it.

My suggestion is to leave it as a parameter to `add_clauses`, but remove the documentation instead of the implementation (or, add a warning to the documentation). If you want to call `CMSat::new_vars(...)`, one of the next things you're going to do is bulk-add clauses via `add_clauses`, so my 2c. is that I'd rather add it here than as a separate (documented) method call. 

~Also, if we remove it and replace it with a function call, we'll likely break the Conda folks.~ I guess I'm the only one using this (or attempting to), so we're fine removing it. Conda folks (who made the PR) don't seem to use it: https://github.com/conda/conda/blob/master/conda/common/logic.py#L227

Anyhow, up to you.

msoos:
Hmm... let's pospone this a bit. I am under a lot of time pressure right now. Can we please come back to this in a week or so? I'll think about it in the meanwhile :) Sorry, I need a clear head and I am super tired unfortunately :( I hope you understand! I see your point and I think you are most likely correct, but as usual, I am afraid of misuse :) Please send me a ping in a week or so when there'll be less deadlines :)

cipherboy:
Sure, no problem. g'luck on the deadlines! :)

PR #555, opened 2019-05-20
Modify the maxTime value to be based off cpuTime (process time) rather than 'solver time', to correctly match the subsequent uses later on

andrewvaughanj:
In the core of CMS (e.g., `Solver::iterate_until_solved`), the value `conf.maxTime` is checked against `cpuTime`, rather than `s.get_stats().cpu_time`.

This PR modifies the setting of `conf.maxTime` to calculate an offset from `cpuTime` to correctly match its later uses.

### DIFF
--- a/src/cryptominisat.cpp
+++ b/src/cryptominisat.cpp
@@ -1,5 +1,6 @@
 /******************************************
 Copyright (c) 2016, Mate Soos
+              2019, Andrew V. Jones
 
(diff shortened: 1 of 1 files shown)


msoos:
Thanks, this is the correct change :)

msoos:
Also, thanks for debugging! Good catch!

PR #556, opened 2019-05-24
Saving glue of unminimised clause in SQL

arijits17:
Created a variable old_glue in handle_conflict. It passes through function calls and reaches dump_clause_stats to get stored in clauseStats table in the database.

### DIFF
--- a/src/prober.cpp
+++ b/src/prober.cpp
@@ -757,7 +757,7 @@ bool Prober::propagate(Lit& failed)
         PropBy confl = solver->propagate<true>();
         if (!confl.isNULL()) {
             uint32_t  glue;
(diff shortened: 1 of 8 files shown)


src/prober.cpp
```diff
@@ -757,11 +757,13 @@ bool Prober::propagate(Lit& failed)
         PropBy confl = solver->propagate<true>();
         if (!confl.isNULL()) {
             uint32_t  glue;
+            uint32_t  old_glue; //Arijit : Not sure whether this is true either
```
msoos: Please remove this comment. It's not needed.

arijits17:
I've removed the comment. Thanks. :)

msoos:
Thanks! Merging :)

PR #566, opened 2019-07-05
Remove Solver_members definition

cipherboy:
The contents of `Solver_members` seem to be left over from a different
time and seem unrelated to CryptoMiniSat. Remove them. In the
`pycryptosat_SolverType` struct, set the `tp_members` pointer to `NULL`.


msoos:
Thanks for all 3, now I understand! Thanks a lot!

PR #583, opened 2019-10-30
Fix extra semicolons in various places and enable the corresponding warning

horenmar:
As part of some different work, I played with warnings that are currently not in use by CMSat and noticed that `-Wextra-semi-stmt` leads to 131 extra warnings, most of which were caused by 3 instances in header files that are then included by many different TUs.


msoos:
Thanks, nice!

PR #586, opened 2019-10-30
Fixes for set_max_confl and set_max_time

horenmar:
The first commit cleans up `set_max_time` and there shouldn't be anything controversial in it.

The second commit cleans up `set_max_confl` and is less a serious commit and more of a suggestion -- the original overflow check invoked UB and every serious compiler would remove it with optimizations enabled. I do not believe that either `int64_t` or `int32_t` (a.k.a `long` on some platforms) can actually overflow in use, but users can use it in a wrong manner that would lead to overflow. The check could be somewhat simplified, if `s->conf.max_confl` was made into an `int64_t`, and `s->get_stats().conflStats.numConflicts` was assumed to always fit within range of `int64_t` (which should be true in practice, as it is not a user-modifiable counter).

### DIFF
--- a/src/cryptominisat.cpp
+++ b/src/cryptominisat.cpp
@@ -505,15 +505,20 @@ DLL_PUBLIC void SATSolver::set_max_time(double max_time)
 
 DLL_PUBLIC void SATSolver::set_max_confl(int64_t max_confl)
 {
-  for (size_t i = 0; i < data->solvers.size(); ++i) {
-    Solver& s = *data->solvers[i];
-    if (max_confl >= 0) {
-      s.conf.max_confl = s.get_stats().conflStats.numConflicts + max_confl;
+  assert(max_confl >= 0 && "Cannot set negative limit on conflicts");
 
-      //don't allow for overflow
-      if (s.conf.max_confl < max_confl)
-          s.conf.max_confl = max_confl;
-    }
+  for (Solver* s : data->solvers) {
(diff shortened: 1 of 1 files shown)


src/cryptominisat.cpp
```diff
@@ -494,32 +494,34 @@ static bool actually_add_clauses_to_threads(CMSatPrivateData* data)
...
-      // as an offset from `cpuTime`.
-      s.conf.maxTime = cpuTime() + max_time;
-
-      //don't allow for overflow
-      if (s.conf.maxTime < max_time)
-          s.conf.maxTime = max_time;
-    }
+  if (max_time < 0) {
+    return;
+  }
```
msoos: Please assert fail. It should be not allowed. Silent failure is a bad idea, people will be chasing their tails. Thanks!

src/cryptominisat.cpp
```diff
@@ -494,32 +494,34 @@ static bool actually_add_clauses_to_threads(CMSatPrivateData* data)
...
 
 DLL_PUBLIC void SATSolver::set_max_confl(int64_t max_confl)
 {
-  for (size_t i = 0; i < data->solvers.size(); ++i) {
-    Solver& s = *data->solvers[i];
-    if (max_confl >= 0) {
-      s.conf.max_confl = s.get_stats().conflStats.numConflicts + max_confl;
+  if (max_confl < 0) {
+    return;
+  }
```
msoos: Again, please fail loudly, i.e. assert fail. Thanks!

msoos:
Thanks a lot, looking great!

PR #588, opened 2019-11-08
Remove 'register' storage specifier from variable declarations in MT

horenmar:
The `register` keyword as a storage specifier keyword has been
deprecated in C++11 and C++17 actually removes it from the language
completely, pending future reuse.

The rationale is that it is a weak hint that is pretty much ignored
by all major modern compilers, and the effect of the specifier is
implicit in C++ anyway.

This also fixes all of the `-Wdeprecated-register` instances in CMSat.


msoos:
Thanks!

PR #590, opened 2019-11-17
Don't use bool as a number

### LINKED ISSUE #589: WalkSAT uses bools weirdly
I noticed that `walksat.cpp` triggers `C4804`, which is "'/': unsafe use of type 'bool' in operation". After closer inspection, this code does not make sense:

```cpp
    if (found_solution) {
        cout << "average flips per assign (over all runs) = " <<
               ((double)totalflip) / found_solution << endl;
        cout << "average seconds per assign (over all runs) = " <<
               (((double)totalflip) / found_solution) * seconds_per_flip << endl;
        cout << "mean flips until assign = " << mean_x << endl;
        cout << "mean seconds until assign = " << mean_x * seconds_per_flip << endl;
        cout << "mean restarts until assign = " << mean_r << endl;
    }
```

Because `found_solution` is guaranteed to be 1 in this block, the calculations can be removed completely (and leaving them in is in fact misleading). Note that the source of `mean_x` and `mean_r` commit the same sin, as they use `found_solution` in a context where it is also guaranteed to be `true` and thus equal to 1.

horenmar:
As it turns out, there were significant changes to `walksat.cpp` between the version we use (latest releaes), and current master. However, some places still used `found_solution` as a number, rather than as a boolean variable, so this fixes (most) of them.

Also removed a variable that was written to, but never read.

Fixes #589

### DIFF
--- a/src/walksat.cpp
+++ b/src/walksat.cpp
@@ -679,7 +679,6 @@ void WalkSAT::update_and_print_statistics_end_try()
         integer_sum_x += x;
         sum_x = (double)integer_sum_x;
         sum_r += r;
-        mean_r = ((double)sum_r) / (double)found_solution;
         x = 0;
         r = 0;
     }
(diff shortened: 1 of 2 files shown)


horenmar:
Note that `flips_this_solution` and `f` are also unused, but I left them in in case they are part of ongoing refactoring.

PR #591, opened 2019-11-17
Prevent an extra copy when iterating through map

horenmar:
The problem is that the entry type of a `std::map<K, V>` is
`pair<const K, V>`, rather than `pair<K, V>`. Because the latter
is constructible from the former via a copy constructor, the old
code compiled, but made an extra copy during iteration.

### DIFF
--- a/src/varreplacer.cpp
+++ b/src/varreplacer.cpp
@@ -1252,8 +1252,7 @@ void VarReplacer::save_state(SimpleOutFile& f) const
     f.put_uint32_t(replacedVars);
 
     f.put_uint32_t(reverseTable.size());
-    for(const std::pair<uint32_t, vector<uint32_t> >& elem: reverseTable)
-    {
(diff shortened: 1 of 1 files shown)


msoos:
There is no copy here, but it's cleaner this way anyway, thanks!

horenmar:
There was an extra copy in the loop, in fact this exact pattern is often used as the poster child for why `auto` and type inference in general is good.

The reason for the copy is that there is a type mismatch between the iteration variable, defined as `std::pair<uint32_t, std::vector<uint32_t>> const&`, and the `const_reference` type of the `std::map`  is `std::pair<const uint32_t, std::vector<uint32_t>> const&`. These two types are convertible, but to do so requires making a copy of the pair -- and the second element of this pair is a `std::vector`.

msoos:
Wow! What a catch! Thanks for explaining!

Mate

PR #594, opened 2019-11-18
Reduce weak vtables

horenmar:
This PR sweets out most instances of weak vtables.

A class has weak vtable when it has no out-of-line definitions of virtual functions. This means that the compiler cannot pick an authoritative TU to contain the class's vtable and RTTI structures. Because of this, the compiler has to emit all of this in every TU that touch the header, bloating the size of object files and increasing compilation times, and link times (because the linker has to deduplicate the vtables and RTTI structures).


msoos:
Niiiice, thank you, I had no idea how to do this!

PR #596, opened 2019-12-04
Extend C API with set_verbosity and print_stats

deiruch:
Does what the title says. cmsat_set_verbosity is included in the C API.


msoos:
Thanks so much! A ton better :) Thanks for caring!

PR #606, opened 2020-02-10
Master rust improvements

berke:
- Provide access to set_max_time() to the C and Rust interfaces.
- In Rust interface, derive Debug for Lit and Lbool.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -17,7 +17,7 @@ pub const MAX_NUM_VARS: size_t = (1 << 28) - 1;
 enum SATSolver {} // opaque pointer
 
 #[repr(C)]
-#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug)]
 pub struct Lit(u32);
 impl Lit {
     /// Returns None if var >= 1 << 31, but you should not rely on var >= MAX_NUM_VARS
@@ -46,7 +46,7 @@ impl std::ops::Not for Lit {
 }
 
 #[repr(u8)]
-#[derive(PartialEq, Eq, Clone, Copy)]
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
 pub enum Lbool {
     True = 0,
     False = 1,
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -84,6 +84,7 @@ extern "C" {
     fn cmsat_get_model(this: *const SATSolver) -> slice_from_c<Lbool>;
     fn cmsat_get_conflict(this: *const SATSolver) -> slice_from_c<Lit>;
     fn cmsat_set_num_threads(this: *mut SATSolver, n: u32);
+    fn cmsat_set_max_time(this: *mut SATSolver, max_time: f64);
 }
 
 pub struct Solver(*mut SATSolver);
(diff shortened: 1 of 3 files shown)


msoos:
Actually, can you check if this could be merged to msoos/cryptominisat-rs?
I have a new repo with a more good stuff! I think it's easier to build with
Cargo too! Could you please check? If not I'll then merge it myself. Sorry
I wanted to deactivate this one and give a nice notice in the README but I
was waiting for feedback from the original author. Let me know what you
think!

https://github.com/msoos/cryptominisat-rs

msoos:
OK, I hand-merged it to both cryptominisat and cryptominisat-rs :) I will now deleting this rust binding from the main "cryptominisat" repo and the official one will be at:

https://github.com/msoos/cryptominisat-rs

I hope that works for you, right? It's super-easy to use it there, see how caqe uses it:

https://github.com/msoos/caqe/blob/master/Cargo.toml#L8

It's kinda cool, no? So I am removing this subdirectory and instead redirect people to that repo. Would that be OK for you?

Thanks,

Mate

msoos:
Also, I gave you credit in both commits:

https://github.com/msoos/cryptominisat/commit/9af568a69a5da5e11805176926b8db82b08eea25
https://github.com/msoos/cryptominisat-rs/commit/da20b3ac4456c9bfe5153a2e94c1c8ee45356cf7

:)

berke:
Hi, thanks for merging the changes. 

I do have trouble compiling my code with
`cryptominisat = { git = "https://github.com/msoos/cryptominisat-rs", branch= "master" }`
though, it doesn't find the `set_max_time` function at link time:
`cryptominisat.2lxmnjk4-cgu.1:(.text._ZN13cryptominisat6Solver12set_max_time17h65acad2e8d3e0966E+0x5): undefined reference to `cmsat_set_max_time

My code is pretty simple, it simply calls `set_max_time` on a solver.  It compiles with my modifications (before you split out the Rust bindings into its own repo.)  I'm not investigating further in case you know what the issue might be.

msoos:
Ah sorry, I'll check and fix!

Mate

msoos:
I can reproduce this bug. But I have no idea what's going on here. It looks exactly like any other function... are you sure it worked for you before? I literally copy-pasted your code, I double and triple-checked and it's the same thing you wrote. Are you sure your code worked before? There seems to be something off here... Do you know if you could investigate perhaps?

msoos:
I have also added a test to cryptominsiat-rs, so `cargo test` will now fail with the same bug:

```
[....]
 in function `cryptominisat::Solver::set_max_time':
          /[...]/cms-rust/src/lib.rs:178: undefined reference to `cmsat_set_max_time'
          collect2: error: ld returned 1 exit status
```

berke:
Hi Mate,

The new `build.rs` is different as it also builds cryptominisat.  Uncommenting

`println!("cargo:rustc-flags=-L {}/lib", dst.display());`

fixes the issue.

BTW in the README, you should add a line or two reminding the user to git submodule init & update.

Cheers

msoos:
Hi!

Actually, it doesn't solve the issue for me... if I uncomment it, it's still the same :S "cargo test" still fails :( 

Good point on adding info on submodule init&update, I added it! Thanks! Do you have any idea how to fix the `cargo test`, though? Any help would be much appreciated!

msoos:
Hi! I think all of this has been fixed now. Can you please check? The Rust binding is now here: https://github.com/msoos/cryptominisat-rs/ Let me know if it works for you too! Cheers,

Mate

PR #607, opened 2020-03-04
Use absolute paths to libraries on cryptominisat5Config.cmake, this i…

arrowd:
…s what CMake expects. Fixes using cryptominisat when installing into unusual prefix.


msoos:
Wait, this broke all my builds :( My builds depend on the built but uninstalled library being registered into CMake's repository of built projects which then automatically resolves it to the right place during build time. Here are two projects that broke:

```
Scanning dependencies of target stp
make[2]: Leaving directory '/home/soos/development/smt/stp/build'
make[2]: Entering directory '/home/soos/development/smt/stp/build'
make[2]: *** No rule to make target '/usr/local/lib/libcryptominisat5.so', needed by 'lib/libstp.so.2.3'.  Stop.
make[2]: Leaving directory '/home/soos/development/smt/stp/build'
make[1]: *** [CMakeFiles/Makefile2:753: lib/CMakeFiles/stp.dir/all] Error 2
make[1]: Leaving directory '/home/soos/development/smt/stp/build'
```

and:

```
[ 98%] Built target MV
make -f src/CMakeFiles/bosphorus.dir/build.make src/CMakeFiles/bosphorus.dir/depend
make[2]: Entering directory '/home/soos/development/bosphorus/build'
cd /home/soos/development/bosphorus/build && /usr/bin/cmake -E cmake_depends "Unix Makefiles" /home/soos/development/bosphorus /home/soos/development/bosphorus/src /home/soos/development/bosphorus/build /home/soos/development/bosphorus/build/src /home/soos/development/bosphorus/build/src/CMakeFiles/bosphorus.dir/DependInfo.cmake --color=
make[2]: Leaving directory '/home/soos/development/bosphorus/build'
make -f src/CMakeFiles/bosphorus.dir/build.make src/CMakeFiles/bosphorus.dir/build
make[2]: Entering directory '/home/soos/development/bosphorus/build'
make[2]: *** No rule to make target '/usr/local/lib/libcryptominisat5.so', needed by 'bosphorus'.  Stop.
make[2]: Leaving directory '/home/soos/development/bosphorus/build'
make[1]: *** [CMakeFiles/Makefile2:460: src/CMakeFiles/bosphorus.dir/all] Error 2
make[1]: Leaving directory '/home/soos/development/bosphorus/build'
```

So I must revert this change. Are you sure this is how this is meant to be? If so, can you please point out to me what is wrong with the build system of e.g. Bosphorus so it wouldn't break as per above:

https://github.com/meelgroup/bosphorus

In the meanwhile, I must revert this change as it breaks all my builds :( Sorry, maybe my other projects are incorrectly set up, though they all seem to have been fine so far, but perhaps they are wrong. Can you please look into this? Thanks!

arrowd:
This change is exactly for `stp`.

When `stp` is built without this patch, its `STPTargets.cmake` ends up with `/usr/local/lib/libminisat.so;cryptominisat5`, which turns into `-lcryptominisat5` flag during linking. But there is no `-L/usr/local/lib` added, so I get a error when I try to link to STP and it pulls cryptominisat in.

How do you build STP? What are the value for `CRYPTOMINISAT5_LIBRARIES` in your case?

arrowd:
Hum. At least bosphorus' code seems sane. Maybe my change is wrong indeed. I'll look into that more.

msoos:
Thanks! Again, you may well be right that *something* is off :D It may well be that both CMS and STP are off :S It'd be great if you looked into it! Thanks, much appreciated! :)

arrowd:
This looks like transitive dependency problem. Here are my findings:

I'm building https://github.com/klee/klee with STP support and takes both `/usr/local/lib/libminisat.so` and `cryptominisat5` libraries from `STPTargets.cmake` file. But since `cryptominisat5` isn't an absolute path, it gets translated to `-lcryptominisat5` and this makes the build fail.

However, adding `find_package(cryptominisat5 CONFIG)` line to KLEE's CMakeLists.txt fixes the issue. This happens because this line causes CMake to import `cryptominisat5Targets.cmake` just like `find_package(STP CONFIG)` imports STP targets.

Now, the question is - should the application project search for all transitive dependencies, or it is STP CMake machinery should be fixed? Maybe `STPConfig.cmake` requires `find_package(cryptominisat5 CONFIG)` line?

msoos:
Thanks for digging! I am happy to change STP to have `find_package(cryptominisat5 CONFIG)` -- can you please verify that it fixes the issue? Thanks!

arrowd:
Like I said,

> However, adding find_package(cryptominisat5 CONFIG) line to KLEE's CMakeLists.txt fixes the issue.

arrowd:
Ah, you mean add it into `STPConfig.cmake`? Let me try that.

arrowd:
Yep, it does work too. Please, add `find_package(cryptominisat5 CONFIG)` to `STPConfig.cmake.in`.

msoos:
Cool! I have never seen a cmake Config file with a find_package in it, though. So let me ask someone on the STP list :) I am closing this now, adding your change as a merge request to STP.

msoos:
Please comment on:

https://github.com/msoos/cryptominisat/pull/607

and let me know if this is how you wanted it to be merged :)

PR #621, opened 2020-07-09
fix build on nix-darwin

Mic92:
in the nix build environment we don't need to explicitly pass
the osx sdk path. In fact it actually breaks the build.


Mic92:
https://github.com/NixOS/nixpkgs/pull/92681

Issue #625, opened 2020-07-31
Support for mingw32


ykazakov:
### Description of the problem

I used to compile windows binaries for cryptominisat using a mingw32 cross-compiler (on MacOS). I observe that for the latest release 5.8.0 this does not work anymore. Is minigw officially supported?

### My environment

- MacOSX 10.15.6
- mingw-w64 7.0.0_2 (homebrew)

### Steps to reproduce

1. Install a `mingw-w64` cross-compiler, e.g.:
```
brew install mingw-w64
```
(on linux there are also mingw-w64 packages)

2. Create `build-ming32.sh` with the following content:

```
#!/bin/bash

set -e
set -x

rm -rf build
mkdir build

export CC=x86_64-w64-mingw32-gcc
export CXX=x86_64-w64-mingw32-g++

(
cd build || exit
cmake -DENABLE_PYTHON_INTERFACE=OFF -DNOVALGRIND=ON -DNOZLIB=ON -DONLY_SIMPLE=ON -DSTATICCOMPILE=ON -DIPASIR=ON -MIT=ON -DCMAKE_BUILD_TYPE=Release -D CMAKE_SYSTEM_NAME=Windows $@ ..
make -j4
)
```

3. Compile cryptominisat

```
$ wget https://github.com/msoos/cryptominisat/archive/5.8.0.tar.gz
$ tar xf 5.8.0.tar.gz
$ cd cryptominisat-5.8.0
$ sh ../build-ming32.sh 
```

### Expected behaviour

This used to compile fine up until version 5.7.1

### Actual behaviour

I get the following compile errors due to `posix_memalign` not available under mingw32

```
In file included from /Users/demo/cryptominisat/cryptominisat-5.8.0/src/gaussian.h:39,
                 from /Users/demo/cryptominisat/cryptominisat-5.8.0/src/clauseallocator.cpp:38:
/Users/demo/cryptominisat/cryptominisat-5.8.0/src/packedmatrix.h: In member function ‘void CMSat::PackedMatrix::resize(uint32_t, uint32_t)’:
/Users/demo/cryptominisat/cryptominisat-5.8.0/src/packedmatrix.h:70:13: error: ‘posix_memalign’ was not declared in this scope
   70 |             posix_memalign((void**)&mp, 16,  size);
      |             ^~~~~~~~~~~~~~
[...]

PR #626, opened 2020-08-01
A patch to address #625

ykazakov:
I have also added a travis option to test a build using a mingw32 cross-compiler on linux, but the travis script also builds other things like drat-trim, and I don't know how to prevent this. The tests are also disabled since it is not possible to run the produced exe file on linux. Ideally, a mingw32 build should run on windows, but I do not have a windows machine to test the setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,6 +97,18 @@ matrix:
       env:
         - CMS_CONFIG=STATIC
 
+    - os: linux
+      dist: bionic
+      compiler: gcc
+      python: "3.7"
+      addons:
+         apt:
+            packages: mingw-w64
+      env:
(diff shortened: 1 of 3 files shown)


msoos:
Oh wow, thanks so much! I love that you added tests, and it caught some issues. Really great stuff, thank you! I'll merge this now and fix the drat-trim compilation issue. Thanks again,

Mate

ykazakov:
You are welcomed! There is also a failed travis build with clang and python 2.7. Not sure why is it failing. I don't think I changed anything relevant.

msoos:
Yep, thanks! It's due to a different (floating point) bug I just fixed :)
Cheers!

Mate

PR #627, opened 2020-08-30
Bugfixes for ipasir_val() and ipasir_assume()

fkutzner:
This PR fixes two issues with the IPASIR implementation:

* **`ipasir_val` for negative literals:** Currently, `ipasir_val(s, lit)` returns `-lit` if the variable `v` of `lit` is assigned `false`, `lit` if `v` is assigned `true` and 0 otherwise. However, the IPASIR spec requires checking the assignment of the literal, not that of the variable. The current implementation works fine for positive literals, but returns invalid values for negative ones (flipping `lit` when the variable of `lit` is assigned `false` even when `lit` < 0).
**New behavior:** `ipasir_val(s, lit)` returns `-lit` if `lit` is assigned `false`; `lit` if `lit` is assigned `true` and 0 otherwise. 

* **`ipasir_assume` for literals not occurring in any clause:** Cryptominisat refuses assumptions not occurring in any clause added up to the next `ipasir_solve()` call. There is no such restriction in the IPASIR spec on the literals that can be passed to `ipasir_assume()`, though.
**New behavior:** `ipasir_assume(s, lit)` creates new variables as needed.

I made a single PR for these two issues to avoid merge conflicts in ipasir_test.cpp, but I could split this one if you prefer separate PRs.

### DIFF
--- a/src/ipasir.cpp
+++ b/src/ipasir.cpp
@@ -92,6 +92,17 @@ DLL_PUBLIC void ipasir_release (void * solver)
     delete s;
 }
 
+namespace
+{
+void ensure_var_created(MySolver& s, Lit lit)
+{
+    if (lit.var() >= s.solver->nVars()) {
+        const uint32_t toadd = lit.var() - s.solver->nVars() + 1;
+        s.solver->new_vars(toadd);
+    }
+}
+}
+
 /**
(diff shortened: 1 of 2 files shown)


msoos:
Wow, that's great, thank you, great findings! Merging :)

PR #642, opened 2021-03-05
Quote some flags in python/CMakeLists.txt

aquohn:
CMake complained about COMPARE needing 4 arguments when I tried to generate using MSYS; looks like the problem is that one of the Python flags is an empty string and CMake thinks the argument is missing. Quoting the flags fixes it.


msoos:
Ah, thanks, great work, merging! :)

Issue #650, opened 2021-04-28
Support cross-compilation by removing check_float_precision


Wimmerer:
Hi! I'm trying to cross-compile CryptoMiniSat for a large number of different environments. Unfortunately the CheckFloatPrecision.cmake file gets in the way of doing that on Linux. I believe it's because it's trying to run a compiled file on the target, but obviously that's not possible in cross-compilation.

What is the exact purpose of this check, and can it be safely removed? I've removed it myself in testing, and everything builds just fine, but I could be silently creating a lot of problems down the line. Running cmake twice works (the dummy values are cached?) or just removing that section of CMakeLists.txt.

Thanks!

andrewvaughanj:
It seems that all this check does is to set `HAVE__FPU_SETCW` and, if that's set, it enables `YALSAT_FPU` -- however, if you're commenting out that check, then `YALSAT_FPU` won't be enabled, so you won't be building with it. I presume that CMS functions fine without `YALSAT_FPU`.

Maybe CMS needs something like:

- https://github.com/getdnsapi/getdns/pull/472/commits/320dd098a8d4d936f439348821e2d7cb707503d5

to avoid having these kind of issues when you're cross-compiling.

Maybe make a PR if this solves it for you?

Wimmerer:
That does work for me. I'll put in a PR this afternoon, although I'd like to know what is lost by removing yalsat. I'll read the Yalsat + CMS paper.

andrewvaughanj:
My understanding of YalSAT is that it is similar to CCAnr, which @msoos describes very briefly [here](https://www.msoos.org/2020/07/cryptominisat-5-8-0-released/) (see: "CCAnr Local Search Solver").

Wimmerer:
This is also related to cross-compiling so I'll note it here to avoid polluting the issues: On musl platforms we get a `/workspace/srcdir/cryptominisat/src/main_exe.cpp:34:5: error: ‘feenableexcept’ was not declared in this scope`, which is caused by a block of code in a `#if defined(__GNUC__) && defined(__linux__)` block. Should this be `(__GLIBC__)`?

andrewvaughanj:
It probably should be `_GNU_SOURCE` based on https://linux.die.net/man/3/feenableexcept

msoos:
[...]

PR #651, opened 2021-04-29
Switch __GNUC__ to __GLIBC__ to support musl

Wimmerer:
Resolves second issue discussed in #650 where musl wouldn't compile since `feenableexcept` isn't available there. Attempted to use _GNU_SOURCE as mentioned by @andrewvaughanj but was not successful. 

Used this method to fix in downstream here: https://github.com/JuliaPackaging/Yggdrasil/pull/2911 where I'm packaging CMS using https://binarybuilder.org/ for Julia!

### DIFF
--- a/src/main_exe.cpp
+++ b/src/main_exe.cpp
@@ -27,7 +27,7 @@ THE SOFTWARE.
 
 int main(int argc, char** argv)
 {
-    #if defined(__GNUC__) && defined(__linux__)
+    #if defined(__GLIBC__) && defined(__linux__)
     feenableexcept(FE_INVALID   |
                    FE_DIVBYZERO |
                    FE_OVERFLOW


andrewvaughanj:
Maybe this would be better to be a CMake-based check anyway, rather than relying on `#defines` like this (given my "guess" was wrong)?

Something like `USE_FEENABLEEXCEPT` maybe?

andrewvaughanj:
Something like https://gitlab.kitware.com/vtk/vtk/blob/bea01e6c2ce4c996b463d79d8a686f48219148a2/Common/Core/CMakeLists.txt#L457 maybe?

That’s not a request, just as a suggestion as to what feels like (to me!) the “most correct” solution.

Wimmerer:
I'll put together something like that tonight!

Might I ask why you prefer a cmake check vs defines? I do vanishingly little c/c++ so I don't have a sense for

andrewvaughanj:
So *I* prefer the CMake-approach because it is then flexible to "anywhere" (e.g., `feenableexcept` is exists on [FreeBSD ](https://www.freebsd.org/cgi/man.cgi?query=fenv), which has its own non-`glibc` libc, and your check will reject that).

Just to be clear: your suggestion is likely to be better than what's there now (e.g., the current code will also reject FreeBSD), but if you're changing it, it would probably be better to change it once-and-for-all vs. just playing what-a-mole.

However, @msoos (the actual maintainer vs. me, an "observer"/few times contributor) might be less worried about being so stringent here :)

andrewvaughanj:
Another argument would be: wouldn't it be nice if CMS automatically could use `feenableexcept` if a future version of `musl` added support for it ... just checking `__GLIBC__` doesn't allow for that.

msoos:
Hey,

First of all, thanks for the very deep discussions and digging! 

Unfortunately, `check_symbol_exists` that's in  https://gitlab.kitware.com/vtk/vtk/blob/bea01e6c2ce4c996b463d79d8a686f48219148a2/Common/Core/CMakeLists.txt#L457 is currently not viable. That's only in `cmake 3.20.2` and I run Arch fully updated, and it hasn't reached Arched stable yet. I mean, this cookie is hot right from the oven :smile: CryptoMiniSat unfortunately needs to compile with a distro from say, 2018. There are quite a few of old systems using CMS.

However, the GLIBC check sounds OK to me. This `feenableexcept` is actually just a way to save my stupid ass when I manage to _again_ do some division by zero/overflow bug somewhere. I usually run clang's ASAN with my fuzzer that can catch floating point bugs too, but sometimes I'm lazy and I don't always run it so this is kinda like a free seatbelt. It's not a necessity actually.

So if you don't mind, I'm gonna merge this :) Let me know if this was a mistake!

Thanks,

Mate

andrewvaughanj:
>  That's only in cmake 3.20.2
>

https://cmake.org/cmake/help/v3.0/module/CheckSymbolExists.html

Looks like it is 3.0?

> Let me know if this was a mistake!
>

Nah, as I mentioned, the change seems better than what was there today, so it certainly isn't wrong :)

msoos:
Oh weird, I have `cmake version 3.20.1` and it was complaining when I copy-pasted that code that the function `check_symbol_exists` does not exist :( Maybe something was wrong with my setup. At least it works now, and I think I managed to fix the yalsat issue as well, so that should help with the cross-compilation :)

andrewvaughanj:
Yeah, I wouldn't stress ... life's too short 💃

PR #658, opened 2021-08-30
Fix Boost linkage

Shatur:
Attempt to fix issue addressed in https://github.com/stp/stp/pull/413

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,7 +248,7 @@ else()
 
     add_compile_options("$<$<CONFIG:DEBUG>:/Od>")
 
-    if (NOT BUILD_SHARED_LIBS)
+    if(STATICCOMPILE)
         # We statically link to reduce dependencies
         foreach(flag_var CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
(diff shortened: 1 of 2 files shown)


msoos:
Beware, the GitHub Actions will break, due to my mess-up. It's just a small issue I'm working to fix.

Shatur:
@msoos, I figured it out. By default CMake uses dynamic runtime but Boost was linked with static runtime. I think this should be a proper fix.
CMake 3.15+ also have [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/git-stage/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html) property. But since the library also supports an older version, I think that this code can be left, I just changed it a little to avoid the warning about overriding parameters.

msoos:
Amazing work! Once it builds on Appveyor, it'll go in :D Thanks a lot, I owe you a drink/beer if we ever meet :)

Mate

andrewvaughanj:
(I'm keeping an eye on this, and I'll wait for this to go in before looking at why STP is unhappy on Windows -- but maybe *this* is the problem as to why STP is unhappy anyway ...)

msoos:
Yaaay, merged! Thank you!

msoos:
We can now do the same for STP hopefully? :)

Shatur:
> Yaaay, merged! Thank you!

Wow, I'm sorry, I'm no checked everything yet. On CI it passes, of course, but I did not have time to make sure that I got rid of the warning about flag overriding. I'll open another PR :)

> We can now do the same for STP hopefully? :)

Sure, will update my PR for STP once we are done here!

PR #659, opened 2021-08-31
MSVC: Fix overriding '/MTd' with '/MT'

Shatur:
Attempt to fix the same issue as in https://github.com/stp/stp/pull/413


Shatur:
@msoos @andrewvaughanj I found the best solution. Should be ready to merge now :)
I also added a fix for warnings that I did in STP.

msoos:
Ah, this looks really great, thanks!! Merging :)

PR #660, opened 2021-09-01
Another approach of static linking

Shatur:
I'm sorry, but I think that the previous approach was was ok (#659) because of [this](https://github.com/stp/stp/pull/413#issuecomment-910490126).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,7 +245,20 @@ else()
     add_compile_options("$<$<CONFIG:DEBUG>:/Od>")
 
     if(STATICCOMPILE)
-        add_compile_options($<$<CONFIG:>:/MT> $<$<CONFIG:Debug>:/MTd> $<$<CONFIG:Release>:/MT>)
+        # We statically link to reduce dependencies
+        foreach(flag_var CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
+            # /MD -- Causes the application to use the multithread-specific
+            #        and DLL-specific version of the run-time library.
+            #        Defines _MT and _DLL and causes the compiler to place
+            #        the library name MSVCRT.lib into the .obj file.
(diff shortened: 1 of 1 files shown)


msoos:
I'm sorry, but are we back where we started? I think I received about 50+ emails about this, and it's basically lead to very little change, if any, if I understand correctly? I understand that this is a lot of work for you, but it's starting to be a lot of work for me too. 

I would like you to clone the repository, sign up for Appveyor, make it work in your repo, and once you are good and happy, create a PR. That way, I don't get 20+ AppVeyor emails and 20+ comment/commit emails about this. Sorry, but I have to concentrate on my work, and this is not helping. I am closing this PR. Please do not open a PR until you feel ready. I am now reverting all changes that have been done regarding the static compilation in Windows. Please create the PR relative to that.

For STP, I'd like you to also significantly reduce the noise by pushing as little as possible, i.e. testing your own changes before sharing with too many people. Thanks for keeping the email noise low in the future,

Mate

CC: @andrewvaughanj  for visibility

Shatur:
Sorry, I didn’t know that this is causing you inconvenience.
The previous approach sets flags correctly correctly (`/MT` for release and `/MTd` for debug). But for unknown reasons it messes MiniSat in STP... I honestly do not understand why it happens.

msoos:
Unfortunately, cmake is complicated. I have worked 8+ years on figuring this thing out so that it works for everyone who gave feedback and for the CI/CD systems that I used. Looking forward to the final PR that makes it all work :) Thanks again for working on this,

Mate

Shatur:
> Unfortunately, cmake is complicated.

Sad but true 😄
Modern CMake simplifies things a little. For example, for my issue there is `MSVC_RUNTIME_LIBRARY` that just works. But in order to maintain compatibility with 3.3, we have to do it manually. And it works, but causes a lot of warnings about overriding `/MD` with `/MT` and uses `/MT` even for debug mode (should be `/MTd`).
This is what I tried to fix :D But it looks like this fix will somehow negatively affect MiniSat in STP. It's weird because CryptoMiniSat links to static Boost libraries correctly. Home I will come with a proper solution soon.

Shatur:
> it seems that your change wasn't just "cleaning things-up" but actually changing the behaviour.

And I do not understand why, because Boost, which is compiled statically, links without problems...

> this is just so that you can use STP/CMS as a sub-project without getting warnings while configuring

Not exactly, on my machine I can reproduce the problem during normal compilation of the repo with `STATICOMPILE` set to `ON` under VS 2019.

PR #666, opened 2022-01-13
Modernize Boost usage

Shatur:
In this PR I modernized Boost usage by using [imported targets](https://cmake.org/cmake/help/latest/module/FindBoost.html#imported-targets). But the feature was introduced in CMake 3.5, so I bumped the version. But KLEE uses [the same version](https://github.com/klee/klee/blob/0ba95edbad26fe70c8132f0731778d94f9609874/CMakeLists.txt#L13), so I hope it's okay :)

Made the same PR for [stp](https://github.com/stp/stp/pull/418).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -754,6 +754,8 @@ endif()
 if(NOT Boost_FOUND)
     set(ONLY_SIMPLE ON)
     message(STATUS "Only building executable with few command-line options because the boost program_options library were not available")
+else()
+    message(STATUS "Boost -- found at library: ${Boost_LIBRARIES}")
 endif()
 
 # -----------------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-cmake_minimum_required(VERSION 3.3 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 11)
 
@@ -754,10 +754,6 @@ endif()
 if(NOT Boost_FOUND)
     set(ONLY_SIMPLE ON)
     message(STATUS "Only building executable with few command-line options because the boost program_options library were not available")
-else()
-    message(STATUS "Boost -- found at library: ${Boost_LIBRARIES}")
-    message(STATUS "Boost -- adding '${Boost_LIBRARY_DIRS}' to link directories")
-    link_directories(${Boost_LIBRARY_DIRS})
 endif()
 
 # -----------------------------------------------------------------------------
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -21,11 +21,6 @@
 include_directories( ${PROJECT_SOURCE_DIR} )
 include_directories( ${BOSPHORUS_INCLUDE_DIRS} )
 
-if(Boost_FOUND)
-    include_directories(${Boost_INCLUDE_DIRS})
-endif()
-
-
 if (NOT WIN32)
     add_cxx_flag_if_supported("-Wno-bitfield-constant-conversion")
     #add_cxx_flag_if_supported("-Wduplicated-cond")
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -754,10 +754,6 @@ endif()
 if(NOT Boost_FOUND)
     set(ONLY_SIMPLE ON)
     message(STATUS "Only building executable with few command-line options because the boost program_options library were not available")
-else()
-    message(STATUS "Boost -- found at library: ${Boost_LIBRARIES}")
-    message(STATUS "Boost -- adding '${Boost_LIBRARY_DIRS}' to link directories")
```
msoos, line 759: I don't understand why this and the previous is removed. This is needed so we know what is being linked. There can be more than one Boost installation.

src/CMakeLists.txt
```diff
@@ -371,7 +366,7 @@ if (NOT ONLY_SIMPLE)
         RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}
         INSTALL_RPATH_USE_LINK_PATH TRUE)
     target_link_libraries(cryptominisat5-bin
-        ${Boost_LIBRARIES}
+        Boost::program_options
```
msoos, line 369: This looks interesting, thanks!

msoos (changes requested):

CMakeLists.txt
Shatur: CMake already prints information about found library. So I thought it's okay to remove it since `link_directories` is no longer necessary. I can put it back if you wish. But usually libraries doesn't print such information manully.

src/CMakeLists.txt
```diff
@@ -21,11 +21,6 @@
 include_directories( ${PROJECT_SOURCE_DIR} )
 include_directories( ${BOSPHORUS_INCLUDE_DIRS} )
 
-if(Boost_FOUND)
-    include_directories(${Boost_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
Shatur: This is how imported targets work. You don't need `include_directories` and `link_directories`. All thanks too `Boost::program_options` :) It automatically adds includes to your project.

msoos:
Sorry, I need to be cautious here. I had a PR a few months ago where someone tried to remove a bunch of outdated things, similar to above, regarding the build system, modernizing it. They had 100% good intentions, and it fixed the build on their end. I merged it, and within days I had people complaining that none of it worked in I multiple different environments. Are you reasonably certain that the above will work in the following cases:

- [ ] Boost is installed to a non-default directory
- [ ] Boost headers are not in a default-included place (e.g. they are under `/opt/devel/`)
- [ ] Boost installed to multiple places, user specifies location via CMake flag which one to use
- [ ] All of the above but on Windows
- [ ] All of the above but on Mac

Sorry, the last person who tried fixing the build (it was Boost related, but to do with static build) made a number of people experience pain, and in the end we had to revert it. In the meanwhile, I lost about 3 working days trying to figure out how to do it "the right way". I feel like I also aged like a year :D

I am happy to merge, but please, for the love of all that is holy, please check it on at least 2 platforms, with at least 2 of the use-cases above.

PS: The removal of the printing of the location needs to be removed from the PR. I don't know where it prints for you without it. Unfortunately, it doesn't print for me, and it can be very confusing when it links the wrong one. So please remove that change, it's literally only printing, you can ignore it if you like.

Shatur:
> someone tried to remove a bunch of outdated things, similar to above, regarding the build system, modernizing it

It was me :smile: But my problem was related to linking. I could not get static linking to work both in my project and on CI. So ended up just linking it dynamically.
But that PR contained also the use of imported targets. I decided to send them separately because you might find it useful.

> I am happy to merge, but please, for the love of all that is holy, please check it on at least 2 platforms, with at least 2 of the use-cases above.

It tested it on Windows with custom Boost location and on Linux with system installation. And CI passed, so I think it should be good.

This change isn't necessary, it just simplifies things a bit, so if you're really worried (BTW, the PR have id #666), you could just not merge it, it's up to you.

Shatur:
> Unfortunately, it doesn't print for me

It's not by default, you need to enable `VERBOSE`. But I put it back as you suggested, you will see the message always as before.

msoos:
Thanks, the changes look better! I'm merging and we'll wait for the positive feedback, hopefully! Sorry, I didn't realize it was you :) Glad you came back and had another go at it, seems like we are at a better place now!

Mate

PR #669, opened 2022-02-12
Add check for NOMPI

capiman:


msoos:
Thank you so much, this has been integrated :) The `NOMPI` option should now exist.

PR #679, opened 2022-04-03
python/setup.py.in: import sysconfig instead of importing it from distutils

xgqt:
setuptools 61 dropped sysconfig module

Ref: https://github.com/stamparm/pcapy-ng/commit/f6ce5248f78ac0a247d76e48cff152f4e3f26482
Bug: https://bugs.gentoo.org/836684

Signed-off-by: Maciej Barć <xgqt@gentoo.org>

### DIFF
--- a/python/setup.py.in
+++ b/python/setup.py.in
@@ -27,7 +27,7 @@ import sys
 import os
 import platform
 from distutils.core import setup, Extension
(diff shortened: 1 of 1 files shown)


msoos:
Thanks so much!!

PR #685, opened 2022-07-28
Add CMake version file

smuenzel:
Add a CMake version file to the default cmake configuration, which enables other cmake packages to use this package when there are version constraints.


msoos:
Thanks so much!

Mate

PR #686, opened 2022-08-16
Add <algorithm> header to assump_test.cpp

maximaximal:
Without this header, std::sort is not included directly in this header file, but transitively over the stp/googletest distribution, in the `include/gtest/internal/gtest-port.h` file, included by `include/gtest/internal/gtest-string.h`, included by `include/gtest/gtest.h`.

This change makes it easier to use a different GTest implementation, which is easier for some package management mechanisms.


msoos:
NICE! Thank you so much!

Mate

PR #690, opened 2022-09-02
Fix Python tests

ozppupbg:
The Python tests did not run with `make test` and did produce errors, if loaded as a module.


ozppupbg:
#691

msoos:
Wow, thank you so much!

PR #708, opened 2023-01-26
Include <cstdint> explicitly where required

rodgert:
These files all assume the definition of types from <cstdint> but were getting those definitions transitively. GCC trunk (GCC13) no longer provides <cstdint> definitions transitively.

### DIFF
--- a/src/alg.h
+++ b/src/alg.h
@@ -20,6 +20,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #ifndef Alg_h
 #define Alg_h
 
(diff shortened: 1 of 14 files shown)


msoos:
Wow, that's really kind of you! I'll need to do this for arjun, AppMC, CMSGen, and UniGen too. I run on Arch so soon things will start failing which is nice. It's kinda hard to know where it's needed without the latest gcc. However, with Arch getting the newest stuff pretty quick, hopefully thing will fail fast on my end so I'll know how to update all these tools.

Thanks again for the work, really good stuff! Let me know if you have any ideas for improvement/etc. I'm always very happy of cool PRs!

Mate

PR #722, opened 2023-06-23
support buffer protocol in Solver.add_clauses

fanosta:
This pull request allows Solver.add_clauses to accept any contiguous buffer according to the [Python buffer protocol](https://docs.python.org/3/c-api/buffer.html) instead of just `array.array`. This possibility was already mentioned in #519.

The main advantage of the buffer protocol is that it can be implemented by third party modules as well. Thus, `add_clauses` will also work with numpy arrays for example. Another advantage is that the C side of the API is much simpler, and thus I was able to remove a bunch of code.

One limitation that remains in the current state of this pull request is that only contiguous buffers, i.e., buffers where the element are contiguous in memory, are supported. Non-contiguous buffers can be created in numpy, for example, if someone indexes only a column of a multidimensional array. If that use-case is desired it would not require too many changes to get it running. (Right now, `PyObject_GetBuffer` will automatically raise a ValueError when used on a non-contiguous buffer.)

Regarding unit-tests, I did not add any new ones but the existing ones for `array.array` cover the changed code and they all passed for me.

### DIFF
--- a/python/src/pycryptosat.cpp
+++ b/python/src/pycryptosat.cpp
@@ -364,28 +364,21 @@ static int _add_clauses_from_array(Solver *self, const size_t array_length, cons
     return 1;
 }
 
-static int _add_clauses_from_buffer_info(Solver *self, PyObject *buffer_info, const size_t itemsize)
+static int _add_clauses_from_buffer(Solver *self, Py_buffer *view)
 {
-    PyObject *py_array_length = PyTuple_GetItem(buffer_info, 1);
-    if (py_array_length == NULL) {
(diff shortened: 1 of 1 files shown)


msoos:
Nice, thanks so much! I'll let the pipelines pass then I'll merge. Then I'll release a new PIP release :)

Thanks again,

Mate

msoos:
Thanks, merged!

fanosta:
thank you very much :)

msoos:
:) I released a new version. The PyPi build & release should be automatic once the pipeline succeeds. Let me know if it goes wrong or something. Otherwise, it should be fine!

Thanks again,

Mate

PR #752, opened 2024-04-17
use tomli instead of toml

### LINKED ISSUE #747: moving from toml to tomli?
[`toml`](https://github.com/msoos/cryptominisat/blame/4c9a6b6b459b7c1381744115ccfe744f1121794b/setup.py#L30) is more or less abandonware now, everybody has moved to `tomli`.
Found this while working on https://github.com/sagemath/sage/issues/34687

Would you mind moving to `tomli` too?

dimpase:
This will close #747


msoos (approved):
NICE, thanks!

PR #744, opened 2024-03-05
IDRUP Support

m-fleury:
Hello,

I have implemented support for IDRUP in Cryptominisat (the paper is available at https://cca.informatik.uni-freiburg.de/papers/FazekasPollittFleuryBiere-MBMV24.pdf). It is a proof format for incremental SAT solving. A checker and a fuzzer are [available](https://github.com/arminbiere/idrup-check) -- although that version does not support CMS yet (not sure when @ArminBiere will update that repo).

The implementation tries to change as little as possible from the structure of the code.

It is not clear to me if you want to include the `ipasir_trace_proof` function to produce idrup proofs. It is not part of the IPASIR interface, but I do not see any other way to create proofs over the IPASIR interactions before [IPASIR2](https://github.com/orgs/ipasir2/discussions) is out.

PR #769, opened 2024-08-28
 IDRUP Support 

m-fleury:
Sorry that it took me way too long to find the time to work on that…

Anyway, this is the rebased and squashed version of https://github.com/msoos/cryptominisat/pull/744

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,7 @@ def gen_modules(version):
                    "src/gaussian.cpp",
                    "src/get_clause_query.cpp",
                    "src/hyperengine.cpp",
+                   "src/idrup.cpp",
                    "src/intree.cpp",
                    "src/lucky.cpp",
                    "src/matrixfinder.cpp",
--- a/src/cnf.cpp
+++ b/src/cnf.cpp
@@ -675,6 +675,14 @@ void CNF::add_frat(FILE* os) {
     frat->set_sqlstats_ptr(sqlStats);
 }
 
+void CNF::add_idrup(FILE* os) {
+    if (frat) delete frat;
+    frat = new IdrupFile<false>(inter_to_outerMain);
+    frat->setFile(os);
+    frat->set_sumconflicts_ptr(&sumConflicts);
+    frat->set_sqlstats_ptr(sqlStats);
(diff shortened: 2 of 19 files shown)


msoos:
Wow, nice, thank you so so much! I'll review and merge today night :)

Mate

msoos:
Wow, thank you so much!!

msoos:
Hey hey!

It' s me from the future with some feedback :) Unfortunately, `idrup` wasn't initialized in the header, but was set as default value : `.default_value(conf.idrup)` so unless `--idrup` was given, it was set to some unknown value. Also, it seems to have assumed that `idrup` is always on, due to calling `conclude_idrup` always I think? So it seemed to have lead to segfault when it wasn't on I think :S Unfortunately, since there is no `fuzz_test.py` changeset to check whether it's correct, I'm not sure it runs/continues to run given my changes over the months :( Maybe I'll write a fuzzer for it. If I don't, I'm afraid it will be broken soon :(

Anyway, now at least FRAT is working in my development version :) I hope the above helps! Cheers and thanks again, idrup is actually really cool as an idea, it just takes time to get it into shape,

Mate

m-fleury:
Do you want me to try to fix it or do you just want to revert the changes?

msoos:
I already fixed it of course. Cheers,

Mate

msoos:
Oh, if you feel like writing a fuzzer for it, that'd be nice! Did you write a fuzzer as part of the paper? In my experience, without a fuzzer, it breaks very easily and hence not really useful in the long term. Is there a fuzzer for it?

Mate

m-fleury:
> In my experience, without a fuzzer, it breaks very easily and hence not really useful in the long term. 

Also my experience.



There is a fuzzer with the checker: https://github.com/arminbiere/idrup-check (the changes to make Cryptominisat are not public).  It is how I tested the code actually. Then I wil try to understand the fuzz_test.py.

msoos:
Ahh I see! I mean, I'd be grateful if you integrated that into fuzz_test.py, but I understand that it's a lot of work. Up to you :)

Mate

PR #775, opened 2024-12-21
Update build instructions to install needed shared libraries

gavinp:
It seems that the default build of cryptominisat5 used shared libraries that weren't installed using these build instructions.

I've added the (scary) minimal stuff I did that makes it work right away.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -37,13 +37,14 @@ cd cadical
 git checkout mate-only-libraries-1.8.0
 ./configure
 make
+sudo cp build/libcadical.so /usr/lib
 cd ..
 
 git clone https://github.com/meelgroup/cadiback
 cd cadiback
 git checkout mate
 ./configure
-make
+make install
 cd ..
 
 git clone https://github.com/msoos/cryptominisat


msoos:
Hi,

Thanks! That's true, the installation is not included. At the same time, to run the executable from where it's at, it's not necessary to install these libraries. Maybe we should make a separate install instruction set? Installing things by default can make people's systems pretty messy, e.g. if they want o later install a different version of the library, or they install it later at a different place (e.g. `/usr/local/lib` instead of `/usr/lib`). So nowadays I am cautious of giving instructions to install things. I have seen colleagues install things to `/usr/lib` and then have trouble with updated libraries colliding.

What do you think?

Thanks again,

Mate

gavinp:
I think your solution is absolutely the best. Installation is indeed not needed; but the executable built in the quick start instructions presumes installed dynamic libraries; one of them from a program that doesn't even install dynamic libraries! Hence my PR including the super scary "cp" to /usr/lib.

Either way, the quick start instructions are materially wrong, and in a way that can confuse many users.

I'll take a few minutes to poke at them and come up with instructions to statically link or something.  Feel free to close this PR if I leave it alone too long, but I should come back with something better.

I'm enjoying looking at the state of open source SAT. I'm curious about the interface between SAT and mixed integer programming solvers.

msoos:
Hi,

Yeah, good point. BTW, it's trivial to build statically! The GitHub actions actually compiles statically:

https://github.com/msoos/cryptominisat/actions/runs/12244747429

Basically, the only difference is to add ` -DSTATICCOMPILE=ON` to the `cmake` of CryptoMiniSat and you'll be good to go:

```
cmake  -DSTATICCOMPILE=ON ..
make
```

Done :) In fact, the reason I use a branch of cadical/cadiback is because by default they don't create both dynamic&static libraries, which would make the above impossible. But anyway, you can also just download it the latest CryptoMiniSat from the Artifacts of GitHub Actions. That will do the trick :)

msoos:
Hi,

I have replaced all of this build instruction stuff with a pointer to the GitHub build, because it's too complicated to explain in text and I don't want to use valuable space for it. I also added a nix build instruction set and a a bunch of build tests for linux, mac, windows for multiple CPU architectures. I hope this fixes the underlying issues.

Mate

Issue #782, opened 2025-04-17
Windows static binary using MSYS2


PolarGoose:
It is possible to produce a statically linked binary of [cryptominisat](https://github.com/msoos/cryptominisat) for Windows using [MSYS2](https://www.msys2.org/).
The resulted binary will not depend on MSYS2 runtime and can be run on any Windows 10/11 x64.

I have created a script for that:
```
#!/usr/bin/env bash
# Run this script from 'ucrt64.exe' MSYS2 shell

set -o xtrace -o errexit -o nounset -o pipefail

readonly currentScriptDir=`dirname "$(realpath -s "${BASH_SOURCE[0]}")"`
readonly buildDir="${currentScriptDir}/build"

# Install prerequisites
pacman --sync --sysupgrade --refresh --noconfirm
pacman --sync --needed --noconfirm \
  mingw-w64-ucrt-x86_64-toolchain base-devel cmake git zlib gmp-devel msys2-runtime-devel ninja

# Create build directory
rm -rf "${buildDir}"
mkdir "${buildDir}"

# Build Cadical
git clone \
  --branch mate-only-libraries-1.8.0 \
  --depth 1 \
  https://github.com/meelgroup/cadical \
  "${buildDir}/cadical"
( cd "${buildDir}/cadical"
./configure -static
make -j `nproc` cadical
)

# Build Cadiback
git clone \
  --branch mate \
  --depth 1 \
  https://github.com/meelgroup/cadiback \
  "${buildDir}/cadiback"
( cd "${buildDir}/cadiback"
./configure
make -j `nproc` libcadiback.a
)

# Build Cryptominisat
git clone \
  --depth 1 \
  https://github.com/msoos/cryptominisat \
  "${buildDir}/cryptominisat"
cmake \
  -S "${buildDir}/cryptominisat" \
  -B "${buildDir}/cryptominisat/build" \
  -G Ninja \
  -D CMAKE_BUILD_TYPE=Release \
  -D GMP_LIBRARY=/ucrt64/lib/libgmp.a \
  -D GMPXX_LIBRARY=/ucrt64/lib/libgmpxx.a \
  -D STATICCOMPILE=ON \
  -D CMAKE_CXX_FLAGS="-w" \
  -D CMAKE_EXE_LINKER_FLAGS="-static -static-libgcc -static-libstdc++"
cmake --build "${buildDir}/cryptominisat/build"

[...]

PR #783, opened 2025-04-18
Github Actions Workflow to build statically linked Windows binary

PolarGoose:
This is a follow-up on the issue: https://github.com/msoos/cryptominisat/issues/782


msoos:
Wow, really nice, thank you!!

PR #819, opened 2026-04-26
C API additions, fix add_bnn_clause_outside edge case

trk20:
Made a few more methods accessible from the C API
- `set_max_confl` 
- `set_timeout_all_calls`
- `interrupt_asap`
- `add_bnn_clause`

Small fix for crash when `add_bnn_clause_outside` is called with `out == lit_Undef`


msoos:
Ahhhhh good catch about the mapping! Merging.

msoos:
Thanks a ton. Sorry for being so effin' late.

PR #826, opened 2026-06-03
Fix BreakID build

### LINKED ISSUE #825: Breakid function name change
Version 3.1.3 of breakid made a backwards-incompatible change: the method `get_sha1_version` was renamed to `get_version_sha1`.  Cryptominisat has not been changed to match, so compilation of `src/cms_breakid.cpp` fails if building with breakid 3.1.3.

msoos:
This should fix #825 and add breakid build to the build.yml


msoos:
@copilot+claude-opus-4.6 review please

PR #827, opened 2026-06-15
Add Windows MSYS2/MinGW wheel builds to python-wheels CI

GregoryMorse:
- Add windows-latest to the build matrix in python-wheels.yml
- Set up MSYS2 with MinGW toolchain (gcc, cmake, ninja, gmp, zlib, pkg-config, ccache)
- Add MinGW to PATH and set CC/CXX for cibuildwheel
- Configure CMAKE_GENERATOR=Ninja and PKG_CONFIG_PATH via CIBW_ENVIRONMENT_WINDOWS
- Add [tool.cibuildwheel.windows] to pyproject.toml with delvewheel repair and Windows-compatible test-command

Windows was already present in build.yml and release.yml but was never added to the Python wheel workflow. This uses the same MSYS2/MinGW approach.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         # Each runner builds for its native architecture.
         # Linux aarch64 uses a native ARM GitHub-hosted runner.
-        os: [ubuntu-latest, ubuntu-24.04-arm, macos-14, macos-15-intel]
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-14, macos-15-intel, windows-latest]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


GregoryMorse:
I am able to build fine on Windows x64 with the latest MSYS2.  Unfortunately I cannot test the release workflow (easily at least) but if enabled, and issues remain, I will solve them.  The last PyPI Windows version still has crash issues so it would be really nice for a fresh release there.

msoos:
Ah nice, thanks! Is this PR AI-generated? Would be nice to disclose. If so, did you review it? Regarding: " The last PyPI Windows version still has crash issues so it would be really nice for a fresh release there." --- ooops.

GregoryMorse:
@msoos Hi and thanks for the approval.  Yes this is AI generated by DeepSeek v4 Pro operating under Copilot.  The local build working fine is not surprising (on Python 3.14 in fact) as you did a recent PR with some work there.  I did look through the changes but as its very GitHub environment specific - I am now going to fix the issues as they come and make some commits until it is fully working.

GregoryMorse:
So I have done as much exhaustive local testing as I can do to try to ensure this will run on GitHub's Windows environment.  Fixed any DLL issues, and silly hardcoded path issues.  Ive run `cibuildwheel` locally as well so I am at about the limit of local testing, the workflow does work.

GregoryMorse:
I'm also fixing now a long pre-existing point of pain on the Python variant, the Ctrl+C support to disrupt potentially very long runs.

GregoryMorse:
The Ctrl+C support across multiple pycryptosat runs is working flawlessly including relinquishing it back to Python when not needed.  The Windows builds also working for standalone and Python, and the cibuildwheel routes.  It might still have GitHub workflow issues but I don't know what I can do besides solve them one by one as you approve the workflow.  Usually it can be solved in a few runs though, as everything is stable and tested.

msoos:
Hi,

I don't like this python lib sigint handler. If you remove it, I'll approve another run for the Windows pypi package to let it build. But this complex signal handler is really unnecessary I think. Looks complex, and hard to debug. Even the normal handler I kinda hate it :) Probably should be removed. But anyway, the python one is definitely not a great idea.

msoos:
I converted this a draft. When you think you are ready, you can mark it "ready for review", I will then review and kick off the pypi build. Thanks!

msoos:
OK. Please clone the repo, make it work there, and then once it's working, please create a PR here :) Otherwise it'll take a week before this is done. Actions are *notoriously* hard to debug, and you'll need to do this *many* times. And I can't approve every time. So just clone, make it work in your own copy and then you don't need approval. Waaaay easier for both you and me :) Thanks! Closing in the meanwhile.

Issue #611, opened 2020-04-06
Ctrl+C doesn't work in pycryptosat


cipherboy:
When a program uses `pycryptosat`, it'll fail to respond to the `Ctrl+C` shortcut because the native component of `pycryptosat`, `SATSolver` from CryptoMiniSat, doesn't invoke Python's signal handler.

Demonstration: 

```python3
from pycryptosat import Solver
import time

class TestSolveTimeLimit:
    def get_clauses(self):
        cls = []
        with open("tests/des16.cnf", "r") as f:
            for line in f:
                line = line.strip()
                if len(line) == 0:
                    continue
                if line[0] == "p":
                    continue
                if line[0] == "c":
                    continue
                line = line.split()
                line = [int(l.strip()) for l in line]
                assert line[-1] == 0
                cls.append(line[:-1])

        return cls

    def test_time(self):
        SAT_TIME_LIMIT = 1
        clauses = self.get_clauses() #returns a few hundred short clauses
        t0 = time.time()
        solver = Solver()
        solver.add_clauses(clauses)
        print("Solving...")
        sat, sol = solver.solve()
        took_time = time.time() - t0

        print(took_time)

test_case = TestSolveTimeLimit()
test_case.test_time()

```

Using `des16.cnf` from [Massacci's](http://disi.unitn.it/~massacci/CryptoSAT/) [des2fml-0.9](https://disi.unitn.it/~massacci/CryptoSAT/des2fml-0.9.tgz) archive for instance.

Expected Output:

```
$ time python3 script.py
^C
$ # script.py exits in a reasonable amount of time after Ctrl+C is hit.
```

Actual Output:
```
$ time python3 script.py
^C
....
$ # script.py exits only after Solver.solve(...) returns with an answer!
```

Note that CMS has to solve the entire instance before it'll return from `pycryptosat.Solver.solve(...)`!

Obligatory output: 

[cmake.txt](https://github.com/msoos/cryptominisat/files/4441224/cmake.txt)
[make.txt](https://github.com/msoos/cryptominisat/files/4441255/make.txt)

msoos:
Thanks! That's a great find! Please use the Pull Request function of GitHub so you can submit a fix. Here is how you can submit a PR:

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests

Looking forward to merging your PR! Cheers,

Mate

GregoryMorse:
[...]

PR #828, opened 2026-06-15
Add python Ctrl+C handler

GregoryMorse:
I think this might be worth a discussion - maybe conditional compilation can placate or alleviate any concerns?  Because on Linux, I understand it is easier to kill processes but if someone is on Windows such as within VSCode, Ctrl+Z, Ctrl+D, etc is totally unresponsive.  So on Windows you have to literally open task manager and kill the process which is kind of unfortunate or a powershell process kill.

Also in Python, perhaps a script does multiple runs and Ctrl+C will have it give up one run and proceed onto the next one.  And Ctrl+C in the intervening period before the next run would cleanly kill the Python process as the handler falls back.

I would say the main justification though is that this is considered the standard de-facto behavior in most libraries.  So in general it is expected, though perhaps this is a special context where aborting is against intentions.  I generally like it during development, of course in production I understand your point, we could even add a flag when initializing the Python library to enable/disable such behavior too.  But I do think it is useful for development purposes especially.

### DIFF
--- a/python/src/pycryptosat.cpp
+++ b/python/src/pycryptosat.cpp
@@ -30,6 +30,7 @@ THE SOFTWARE.
 #include <cassert>
 #include <algorithm>
 #include "../../src/cryptominisat.h"
+#include "../../src/signalcode.h"
 using namespace CMSat;
 
 #define MODULE_NAME "pycryptosat"
@@ -586,10 +587,12 @@ static PyObject* solve(Solver *self, PyObject *args, PyObject *kwds)
         return NULL;
     }
 
+    solverToInterrupt = self->cmsat;
     lbool res;
     Py_BEGIN_ALLOW_THREADS      /* release GIL */
     res = self->cmsat->solve(&assumption_lits);
     Py_END_ALLOW_THREADS
+    solverToInterrupt = nullptr;
 
     self->cmsat->set_verbosity(self->verbose);
     self->cmsat->set_max_time(self->time_limit);
@@ -824,6 +827,9 @@ MODULE_INIT_FUNC(pycryptosat)
         Py_DECREF(m);
         return NULL;
     }
+
+    // Set up Ctrl+C (SIGINT) handler so long solve() calls can be aborted.
+    setup_signal_handler();
 
     return m;
 }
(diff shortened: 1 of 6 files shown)


msoos:
Hi,

Firstly, this still has the windows python github action thing. That should be a separate PR. Secondly, this PR has wayyyy too many comments that sound like an AI wrote it. Did you manually review this PR?

GregoryMorse:
> Hi,
> 
> Firstly, this still has the windows python github action thing. That should be a separate PR. Secondly, this PR has wayyyy too many comments that sound like an AI wrote it. Did you manually review this PR?

You are right I had forgotten to cherry pick the commits here.  Anyway it is good to go now.  I did fully review it - that is why it took 5 commits, I thought it was a Windows issue then I realized it was a Python issue, and then I was revising it so multiple Ctrl+C as well as Ctrl+C outside of pycryptosat after running it still all worked with expected behavior.  So the current code is per my exact guidance and expectations.  I can of course remove the comments if you don't like them.

GregoryMorse:
From https://docs.python.org/3/c-api/exceptions.html: "PyErr_CheckSignals() should be called by long-running C code frequently enough so that the response appears immediate to humans."  So this is a strong suggestion as it follows traditional responsive design rules.

However polling is really not an elegant or efficient solution so we follow the well-known alternative: https://stackoverflow.com/a/53134263.

msoos:
signalcode.cpp pollutes the namespace as-is for the library. It'd need to have everything in a namespace, or it all goes into the global namespace. That needs fixing before this can be approved.

GregoryMorse:
> signalcode.cpp pollutes the namespace as-is for the library. It'd need to have everything in a namespace, or it all goes into the global namespace. That needs fixing before this can be approved.

All taken care of.  It is now in the CMSat namespace.

GregoryMorse:
Fixes #611 which seems to have be inadvertently closed as Ctrl+C was only fixed in the standalone not the Python version

msoos:
OK, opened #832 into which I can push into (I can't push into this), and it needed some cleanup. Closing this and will merge that. It has your commits and your name, so your work is not being erased :)

Thanks!!

PR #829, opened 2026-06-16
Add windows python wheels

GregoryMorse:
- Add windows-latest to the build matrix in python-wheels.yml
- Set up MSYS2 with MinGW toolchain (gcc, cmake, ninja, gmp, zlib, pkg-config, ccache)
- Add MinGW to PATH and set CC/CXX for cibuildwheel
- Configure CMAKE_GENERATOR=Ninja and PKG_CONFIG_PATH via CIBW_ENVIRONMENT_WINDOWS
- Add [tool.cibuildwheel.windows] to pyproject.toml with delvewheel repair and Windows-compatible test-command

Windows was already present in build.yml and release.yml but was never added to the Python wheel workflow. This uses the same MSYS2/MinGW approach.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -681,8 +681,11 @@ if(BUILD_PYTHON_EXTENSION)
 
     target_compile_definitions(pycryptosat PRIVATE
         CMS_FULL_VERSION="${CMS_FULL_VERSION}")
+    # MinGW can't use MSVC .lib to resolve __imp_ stubs.  Undefine
+    # dllimport from Python API macros but keep dllexport for PyInit_.
     target_compile_options(pycryptosat PRIVATE
-        "-D__declspec(x)=")  # prevent __imp_ stubs (MinGW vs MSVC DLLs)
+        "-DPyAPI_FUNC(RTYPE)=RTYPE"
+        "-DPyAPI_DATA(RTYPE)=extern RTYPE")
     target_compile_features(pycryptosat PRIVATE cxx_std_17)
     install(TARGETS pycryptosat DESTINATION . COMPONENT python)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -656,6 +656,9 @@ if(BUILD_PYTHON_EXTENSION)
         set(Python_INCLUDE_DIRS "${_py_base}/include")
         # MinGW can link directly against the DLL.
         set(Python_LIBRARIES "${_py_base}/python3.dll")
+        # Must call find_package with Development (even if non-required)
+        # to make python_add_library available.
+        find_package(Python COMPONENTS Development)
     else()
         find_package(Python REQUIRED COMPONENTS Development.Module)
     endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -651,29 +651,19 @@ if(BUILD_PYTHON_EXTENSION)
             "import sys; print(sys.base_prefix)"
             OUTPUT_VARIABLE _py_base OUTPUT_STRIP_TRAILING_WHITESPACE)
 
+        # Create an IMPORTED target for python3.dll so CMake handles
+        # the linking correctly (resolves __imp_ stubs for MinGW).
+        add_library(python3_shared SHARED IMPORTED)
+        set_target_properties(python3_shared PROPERTIES
+            IMPORTED_LOCATION "${_py_base}/python3.dll"
+            IMPORTED_IMPLIB   "${_py_base}/libs/python3.lib")
+
         add_library(pycryptosat MODULE python/src/pycryptosat.cpp)
         target_include_directories(pycryptosat PRIVATE
             "${_py_base}/include"
             ${PROJECT_SOURCE_DIR}/src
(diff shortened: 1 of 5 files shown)


GregoryMorse:
@msoos so you were right it was a bit brutal to get working - first time I've tried a github workflow with mingw.  In fact I could do it on my branch by adding its name to the branches list in the yml file.  We will need a version bump.  Also I can't test the publish step obviously, often I see a publish-test which essentially does a staging for PyPI though I am not sure it is needed here.

GregoryMorse:
Had some local DLL hell that I have validated is clean in the CI delvewheel version.  I added the publish-test / dry-run to make this maintainer friendly.  Now there is very high confidence a Windows PyPI publish will go successfully.

GregoryMorse:
Fixes #830

src/occsimplifier.cpp
msoos, line 1: This looks completely unacceptable. Whatever this is, cannot go in.

CMakeLists.txt
msoos, line 1: Are we SURE we need this mess? This looks like a huuuuuge mess

msoos:
Please do NOT push here until you have CLONED the repository and then BUILT it there via actions and you are 100% CERTAIN that it will build. I cannot push approve on every try. You can CONE and TEST all by yourself. Please do that. I am closing in the meanwhile. There should be no need for any confidence. It should be FULLY tested and THEN, and ONLY then a PR opened.

src/occsimplifier.cpp
GregoryMorse: First time I have used pycryptosat on Windows with no issues - that picosat failure was a very long standing issue with too many threads.  It is because malloc with MSVCRT is ancient in MinGW and can get corrupted.  I don't think there is any other fix.  This is a crash that has existed on Windows for years.  It is conditionally compiled only for Windows.  The problem is picosat and MinGW.  On Linux the ptmalloc is stable.  It is known if you research it.  Ugly yes.  Tested and absolutely working without crash also yes.

CMakeLists.txt
GregoryMorse: Yes unfortunately.  Locally I didnt.  But in the mingw CI, it is a mess.  I don't disagree.  But mingw is not that common anymore and not well supported and that mess does in fact cleanly work.

msoos:
Why MinGW? This whole thing works under normal Windows MSVC? I think I'll have a go with that. Please do not update or in any other way re-create this PR. I'll do this myself. This looks like a huge mess as-is.

GregoryMorse:
> Please do NOT push here until you have CLONED the repository and then BUILT it there via actions and you are 100% CERTAIN that it will build. I cannot push approve on every try. You can CONE and TEST all by yourself. Please do that. I am closing in the meanwhile. There should be no need for any confidence. It should be FULLY tested and THEN, and ONLY then a PR opened.

I have done exactly that as you had requested so it seems pretty fast to close this PR: https://github.com/GregoryMorse/cryptominisat/actions/runs/27630403931/job/81703534632

GregoryMorse:
> Why MinGW? This whole thing works under normal Windows MSVC? I think I'll have a go with that. Please do not update or in any other way re-create this PR. I'll do this myself. This looks like a huge mess as-is.

The testing requires very large instances with 8 threads to actually experience the race conditions and crashes.  Something I have a harness for but does not exist in the project currently.  My testing has gone above and beyond to find the heap issue

GregoryMorse:
> Why MinGW? This whole thing works under normal Windows MSVC? I think I'll have a go with that. Please do not update or in any other way re-create this PR. I'll do this myself. This looks like a huge mess as-is.

@msoos Why do you advertise MinGW as the correct way if it is not.  I mean I went down this rather annoying rabbit hole that indeed does create a mess thanks to your own instructions: https://github.com/msoos/cryptominisat#building-from-source - it is required because CMake uses gcc and has no MSVC route and would require a significant porting job.  MinGW just works with the current CMake configuration.  But MSVC would avoid this threading fix for sure, it is old MinGW that causes that whole mess, not to mention the build hacks for CI.  It is tested and working and fully confirmed end-to-end though as I have shown in the proof.  MSVC would be simpler and cleaner but require a lot more changes maybe throughout the source code - changes it is unlikely you will want to deal with unless you have a Windows system.

msoos:
OOops! Sorry :) Okay, let me have a go at this :)

GregoryMorse:
> OOops! Sorry :) Okay, let me have a go at this :)

No problem hope this work at least helps you - I have long dreamed of full Windows support.  No crashes is such a relief - I am actively solving problems at scale with it for the first time ever which is really cool.

msoos:
Sorry for being so difficult. I’ll merge your signal thing about python and
then have a go at this :) Thabks for your patience!

GregoryMorse:
> Sorry for being so difficult. I’ll merge your signal thing about python and then have a go at this :) Thabks for your patience!

For me all that is important is we get proper WIndows support.  My fix for the crashes on Windows was alas a hack not the right one.  The problem is pinpointed to how picosat is used and allocator-sensitive code.

Issue #830, opened 2026-06-16
Race condition causes crashes in multithreaded contexts


GregoryMorse:
An `Access violation - code c0000005 (first chance)` occurs with many threads.

The following stack trace was captured in WinDbg (was not easy to pin down, had to use 8 threads and a very real-world scale problem):
```
[0x0]   pycryptosat_cp314_win_amd64!picosat_add+0x55e   0x2640a5ef50   0x7ff99d03c837   
[0x1]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::add_picosat_cls+0xc7   0x2640a5efe0   0x7ff99d03cb11   
[0x2]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::find_irreg_gate+0x1d1   0x2640a5f050   0x7ff99d049066   
[0x3]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::test_elim_and_fill_resolvents+0x416   0x2640a5f110   0x7ff99d0494ca   
[0x4]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::maybe_eliminate+0x11a   0x2640a5f1c0   0x7ff99d04a278   
[0x5]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::eliminate_vars+0xc78   0x2640a5f230   0x7ff99d04cd70   
[0x6]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::execute_simplifier_strategy+0x1660   0x2640a5f3a0   0x7ff99d04d1e7   
[0x7]   pycryptosat_cp314_win_amd64!CMSat::OccSimplifier::simplify+0x317   0x2640a5f7a0   0x7ff99cfc246a   
[0x8]   pycryptosat_cp314_win_amd64!CMSat::Solver::execute_inprocess_strategy+0x15ea   0x2640a5f840   0x7ff99cfc3778   
[0x9]   pycryptosat_cp314_win_amd64!CMSat::Solver::simplify_problem+0x1b8   0x2640a5fb00   0x7ff99cfc6700   
[0xa]   pycryptosat_cp314_win_amd64!CMSat::Solver::iterate_until_solved+0x1c0   0x2640a5fb50   0x7ff99cfcb4e8   
[0xb]   pycryptosat_cp314_win_amd64!CMSat::Solver::solve_with_assumptions+0x298   0x2640a5fbc0   0x7ff99d10cf6f   
[...]

PR #831, opened 2026-06-17
Fix stale PicoSAT literal pointer crash under multi-threaded solve

GregoryMorse:
PicoSAT's inc_max_var() can realloc ps->lits when new variables are encountered mid-clause.  Although fix_added_lits() patches buffered Lit* pointers in ps->added, the realloc is layout-sensitive: under heavy 8-thread allocation pressure the patch can miss a pointer, causing a dereference of a stale/corrupt Lit* in trivial_clause().

Fix: pre-register all variables with picosat_inc_max_var() before any picosat_add(...literal...) opens a clause, so ps->lits is fully grown before any Lit* pointers are buffered.  Added:

- register_lit_to_picovar() — registers a variable with PicoSAT without incrementing picolits_added (used during pre-registration)
- pre_register_picosat_vars() — walks watched clauses and pre- registers all variables via register_lit_to_picovar()
- Called at all three PicoSAT call sites before clause construction
- release_assert in lit_to_picolit() that the variable was pre- registered, so any future pre-pass mismatch is caught immediately rather than manifesting as a rare heap crash

Also: skip red binary watches in pre-registration (matching the filter in add_cls_to_picosat_definable), and reset picosat on added==0 in the definable paths to avoid accumulating unused vars.

### DIFF
--- a/src/occsimplifier.cpp
+++ b/src/occsimplifier.cpp
@@ -1509,19 +1509,19 @@ vector<OrGate> OccSimplifier::recover_or_gates()
     return or_gates;
 }
 
-int OccSimplifier::lit_to_picolit(const Lit l) {
-    picolits_added++;
+void OccSimplifier::register_lit_to_picovar(const Lit l) {
     auto f = var_to_picovar[l.var()];
-    int picolit = 0;
     if (f == 0) {
         int v = picosat_inc_max_var(picosat);
         var_to_picovar[l.var()] = v;
         picovars_used.push_back(l.var());
-        picolit = v * (l.sign() ? -1 : 1);
-    } else {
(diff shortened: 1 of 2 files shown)


GregoryMorse:
Fixes #830

GregoryMorse:
I have now run it for many hours over many various large instances and it is finally observably stable.

I don't think this is Windows only and it is bizarre that it has not been observed previously.  It can be that threads>=8 and realloc on linux are less likely to change allocation addresses.  Picosat depends on realloc and pointer delta adjustments.  Picosat is fragile and too address-sensitive.  The fix here is a clean way around having to worry about it.

msoos (approved):
NICE. This is actually a bug! Thanks for fixing.

PR #837, opened 2026-08-29
CMake options to rename exported CaDiCaL/CadiBack

TrevorHansen:
Hiya Mate, long time! Looks like you've been keeping busy.

A little PR because we don't use CaDiCaL/CadiBack with STP, so it complicates our build for no benefit.

Happy to take improvements on board.


msoos:
CMS nowadays auto-downloads and autobuilds all its dependencies, thanks to the AI-generated wonderful CMakeLists. It also fixed the static build, and it also created a nix build. It's worth taking a shot at doing what I'm doing. It really does work very well. CMakeLists will just download, configure and then as part of `--build`, will build all sub-projects. It's wonderful. It actually works on Linux, Mac, and Windows! It's crazy. Please check it out -- you can simply point the LLM to my repo and tell it to do the same to STP. It will do it in less than 15 minutes and then STP will auto-download and build all dependencies on all platforms, and literally *just work*. I would MUCH rather you did that than doing this -- this looks like a hack. Can you take a shot at it? It will actually make the build system _easier_, I promise. It's crazy good. You can literally build my most complicated system, Ganak, that uses CMS, Arjun, ApproxMC, BreakID, TreeCutter, a bunch of graph libraries, CaDiCal, CadiBack, etc. and just do:

```
        cd project
        mkdir -p build && cd build
        cmake \
        -DENABLE_TESTING=OFF \
        -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
        -DBUILD_SHARED_LIBS=${{ matrix.shared_libs }} \
        -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
        -S ..
```

And it will build on ALL platforms, with BOTH static and dynamic linking (just pass ON/OFF for `matrix.shared_libs`), recursively auto-download all dependencies, configure them all correctly, and build it all. It's absolutely nuts. And it's really magic-free. Please, I ask you to take a look at this first, before  we make this hack. Check out this workflow file that is TRIVIAL and works al ALL platforms:

https://github.com/msoos/cryptominisat/actions/runs/32645444390/workflow

Let me know what you think,

Mate

TrevorHansen:
Completely agree! We've had the same experience with the AIs and CMake files. The AIs are amazingly good at CMake files and making solvers.  Over the last 3 months we've tripled the amount of code & features in STP (see below).

Regards building CMS. I don't know what @aytey tried. But when I tried,  I ran into a wall because  CMS (which of course includes CaDiCaL) got name conflicts when I tried to also link directly with CaDiCaL.

I'll give it over to the AI to try again. For extra challenge, ABC, which we still use for generating the AIG/CNF, also bundles CaDiCaL. Perhaps it can get three copies of CaDiCaL to play nicely. 

Hope you're enjoying the future.

```mermaid
xychart-beta
    title "STP lines of code (total / solver+tools / tests)"
    x-axis ["30 May", "6 Jun", "13 Jun", "20 Jun", "27 Jun", "4 Jul", "11 Jul", "18 Jul", "25 Jul", "1 Aug", "8 Aug", "15 Aug", "22 Aug", "30 Aug"]
    y-axis "Lines" 0 --> 220000
    line [73377, 73377, 73377, 73377, 73377, 73377, 73377, 82342, 92428, 98579, 130302, 157607, 181995, 207271]
    line [58264, 58264, 58264, 58264, 58264, 58264, 58264, 61425, 65877, 68955, 87211, 106930, 121583, 136493]
    line [11218, 11218, 11218, 11218, 11218, 11218, 11218, 17022, 22571, 25632, 39099, 46685, 56420, 66786]
```

msoos:
Nice! Beware, AI is great at writing code, but it can also bloat things up. I let it create a lot of tests and fuzz systems, but lately I have noticed it added a ton of code, but for very little benefit. In fact, it was slower, with 5k extra lines of code :D So you have to be really on top of it. Also, with AI it's now very easy to add fuzzers for each module. I suggest doing that. Also, be  on top of the fuzzer -- it'll add exceptions for bugs into the fuzzer so it doesn't detect them :D It's rather funny sometimes. You really need to read through the fuzz and test codes. Otherwise it'll go haywire. It's also good to test the system's performance regularly, and whatever it added that doesn't make the system substantially faster, should be removed. Otherwise it'll bloat into eternity :D

Regarding CdiCaL --- ohhh, actually, clashes with another version of CaDiCaL is very possible... true. Damn. I should switch to regular CaDiCaL.... Let's see what you get out of it, but I am afraid it might be a problem. Let me know how it goes. If it doesn't work, I'll add (a version of this) PR.

msoos:
Reopening in case there will be conflicts. Let me know how it goes!

TrevorHansen:
> Nice! Beware, AI is great at writing code, but it can also bloat things up. I let it create a lot of tests and fuzz systems, but lately I have noticed it added a ton of code, but for very little benefit. In fact, it was slower, with 5k extra lines of code :D So you have to be really on top of it. Also, with AI it's now very easy to add fuzzers for each module. I suggest doing that. Also, be on top of the fuzzer -- it'll add exceptions for bugs into the fuzzer so it doesn't detect them :D 

Yes! Same experience. My AI disabled the oracle of my fuzzer because it segfaulted sometimes. 😂

I haven't gone through a full cycle yet, but this is how I'm expecting to work:
* Implement new features enabled/disabled by a flag.
* Lots of testing on each new feature to check that it's correct.
* Get the AI to make the new feature fast, with byte-identical output.
* Do fuzzing with lots of combinations of flags, so they can all be enabled/disabled separately.
* Do parameter optimisation over the flags on our benchmark set.
* Delete each unhelpful feature, recording details of it to a report, so we don't try it again.

Is this going to work OK you think?

TrevorHansen:
I've had another go. See the changes. This keep CaDiCaL in the build, but renames it so that it doesn't conflict. A few other things (a) this can only work if CaDiCaL is built by CMS, not re-used from the system, and (b) it needs to exclude the CaDiCaL C-API, which I don't imagine you want to export anyway.

Is this getting closer?

